PR #38, opened 2017-06-19
Include filename in vp

peterdesmet:
It's sometimes useful to know from which file a vp object was generated, so this pull request adds the attribute `filename` to vp object. It contains the filename (path) that was fed to `readvp()`:

```r
> str(vp_files[[1]])

chr "../data/raw/example/seang_vp_20161003T2000Z.h5" # This is new
$ radar     : chr "SE50"
$ datetime  : POSIXct[1:1], format: "2016-10-03 20:00:00"
$ data      :'data.frame':	25 obs. of  16 variables:
...
$ attributes:List of 3
```

As this is a new attribute, it should not affect any existing code.

This functionality already comes in handy in https://github.com/enram/vp-processing, where the filename is used to get a consistent radar id (and datetime): see https://github.com/enram/vp-processing/issues/18.

### DIFF
--- a/R/bioRad.R
+++ b/R/bioRad.R
@@ -318,7 +318,7 @@ readvp = function(filename){
   }
 
   #prepare output
-  output=list(radar=radar,datetime=datetime,data=profile,attributes=list(how=attribs.how,what=attribs.what,where=attribs.where))
+  output=list(filename=filename,radar=radar,datetime=datetime,data=profile,attributes=list(how=attribs.how,what=attribs.what,where=attribs.where))
   class(output) = "vp"
   output
 }


adokter:
@peterdesmet Some ambiguity here what `filename` refers to, it can either be a (raw) polar volume file, or a vertical profile file. I think you now added the vertical profile file name that stores a VP object, but vol2bird in bioRad may also generate a VP object on the fly from a polar volume file.

My suggestion would be to:
* add the polar volume source file as an attribute that is added by vol2bird.
* not store the vertical profile file name, as not all VP objects necessarily have one associated with it.

peterdesmet:
> but vol2bird in bioRad may also generate a VP object on the fly from a polar volume file.

Is that polar volume file then the `filename` passed to `readvp()` or is that vp object created without `readvp()`?

Regarding suggestions:

- Add polar volume source file name: doesn't really help my use case and won't be included in existing vp files, but always good to include for new files.
- I would add vertical profile file name *when available*: we can really use it for further processing of vertical profiles!
- Agree to add it to attributes.

So maybe the format could be:

```
attributes$polarvolume_filename
attributes$vp_filename
```

What do you think?

peterdesmet:
Ping @adokter: this is the pull request we discussed. See https://github.com/adokter/bioRad/pull/38#issuecomment-309683712: how do you want the attribute to be named?

adokter:
Hi @peterdesmet, thanks for reminder on this one. Sounds good to add these attributes. I prefer to go with 
```
attributes$pvol_filename
attributes$vp_filename
```
to keep nomenclature consistent for pvol and vp objects

adokter:
closed by commit dadf73dadac9d2650724811e71c5175209afb382

peterdesmet:
Thanks @adokter! vp-processing is now updated to use this new attribute. To avoid confusion, I'll remove the ENRAM fork of bioRad, so people just find this one. 👌

PR #53, opened 2017-12-08
newly rendered pkgdown website in docs/ folder

stijnvanhoey:
This is the result from a local new `pkgdown::build_site(".")` webstie generation and the inclusion of html files. I propose to merge this PR and check if the rendering works out fine.


stijnvanhoey:
One of the next steps is to provide more structure in the reference section as it is a single long list now. We can merge for now and I'll improve the website handling later on.

adokter:
@stijnvanhoey thanks, the web page is up now!

PR #64, opened 2018-02-14
Fix #63 on date sequence

### LINKED ISSUE #63: start_date - end_date bug in vp_download
The following code:

```r
library(bioRad)

download_vp(
  start_date = "2016-09-25", # September
  end_date = "2016-10-05", # October
  country = c("pt"),
  radar = c("far"),
  localpath = "."
)
```

Is expected to download all data from `far` from the months September and October ([which is available](http://enram.github.io/data-repository/?prefix=pt/far/2016/)). Note: it downloads the data from the whole month, because it uses the monthly zip files.

However, only September (`ptfar201609.zip`) is downloaded. October (`ptfar201610.zip`) is not.

It seems that the bug is caused when the start **day** (`19`) is larger than the end **day** (`05`). 😱  This works fine:

```r
library(bioRad)

download_vp(
  start_date = "2016-09-05", # September, but start day is less or equal then end day
  end_date = "2016-10-05", # October
  country = c("pt"),
  radar = c("far"),
  localpath = "."
)
```

And the bug is consistent:

```r
library(bioRad)

download_vp(
  start_date = "2016-09-25", # September, but start day is larger then end day
  end_date = "2016-11-05", # November
  country = c("pt"),
  radar = c("far"),
  localpath = "."
)
```

Will only download September and October, but not November.

stijnvanhoey:
Quick fix of the issue noticed by @peterdesmet, making sure the `seq` takes into account all involved months.

### DIFF
--- a/R/download_vp.R
+++ b/R/download_vp.R
@@ -41,7 +41,7 @@ download_vp <- function(start_date, end_date, country, radar, localpath = ".") {
 
   # create date range set of potential downloadable zip files (if all data
   # would exist)
-  start <- as_date(start_date, tz = NULL)
+  start <- floor_date(as_date(start_date, tz = NULL), "month")
   end <- as_date(end_date, tz = NULL)
   dates_to_check <- seq(start, end, by = 'months')
 


adokter:
Oh we were fixing it in sync :)

Issue #56, opened 2017-12-12
some functions in NAMESPACE not shown in pkgdown webpage


adokter:
For example `is.vpts` and `dim.vpts` are missing, while `is.ppi` and `dim.ppi` do show up correctly.

stijnvanhoey:
I checked all NAMESPACE functions and the same problem arises for `is.vplist`. What's happening is the wrong rendering towards pkgdown when having the `[.vpts` or `[.vplist` redirected `@rdname` to the summary. On the page itself, it get rendered:
![image](https://user-images.githubusercontent.com/754862/37062176-fd14b0f6-2195-11e8-8e21-da477bcdaa46.png)
but as it is interpreted as a single code block instead enlisting of methods, the differentiation is not taken into account to the overview. I assume for now this is a pkgdown issue we should check.

I first thought to make the `vpts` similar to `ppi`, but actually I do think that for `[.ppi` the `@rdname` [towards `ppi` function](https://github.com/adokter/bioRad/blob/master/R/bioRad.ppi.R#L415) should rather be the [link to the `summary.ppi`](https://github.com/adokter/bioRad/blob/master/R/bioRad.ppi.R#L873) as it is with `vpts` and `vplist`. @adokter, does this make sense? Doing so is off course causing the removal of `is.ppi` and `dim.ppi` from the overview as well. 

One solution would be to not use the `@rdname`, but providing separate pages for each of the functions.

stijnvanhoey:
By providing a separate page for the `[.ppi`, keeping the others (`summary`, `dim` and `is.*`) grouped, we actually have a solution that renders:
- the overview
![image](https://user-images.githubusercontent.com/754862/37063142-ceea62cc-2198-11e8-954e-f34275ca6d1a.png)
- combination of the ref of `ppi`
![image](https://user-images.githubusercontent.com/754862/37063192-e40854c0-2198-11e8-811e-bbe4f364bcda.png)
- the separate docstring for the subset (I now quickly added `subset ppi` to roxygen header)
![image](https://user-images.githubusercontent.com/754862/37063215-f6756b16-2198-11e8-9317-bd0a41fd8452.png)

Would this be an option for `vp`, `vplist` and `ppi`?

adokter:
@stijnvanhoey Thanks, yes I think an extra page for the subsetting isn't that bad, I mainly care about having all the functionality documented. So please go ahead as you suggest

Issue #50, opened 2017-11-17
Split R code over files by function


peterdesmet:
For better readability, maintenance and collaboration, I'd suggest to split [bioRad.R](https://github.com/adokter/bioRad/blob/master/R/bioRad.R) (and the other R files) in logical files: one per function (and its helper functions), with the same name as that function.

@adokter these are all the functions I found, per current file. It's quite the list 😅 At first glance, think it would be better to group these by what they do (e.g. bind, vintegrate) that their data type (vp, vplist).

## bioRad.R

function | remarks | new file | reference section
--- | --- | --- | ---
checkDocker | in NAMESPACE | docker.R | docker interface
updateDocker | in NAMESPACE | docker.R | docker interface
startContainer | | docker.R | 
readOdimProfileData | | hdf5.R
quantityName | | hdf5.R
readvp | in NAMESPACE | vp.R | vertical profile
is.vpfile | in NAMESPACE | vp.R | vertical profile
is.pvolfile | in NAMESPACE | pvol.R | polar volume
h5ODIMobject | in NAMESPACE | hdf5.R | loading data
is.ODIMfile | | hdf5.R
readvp.list | in NAMESPACE | vp.R | loading data
c.vp | in NAMESPACE | vp.R | vertical profile
bind | in NAMESPACE | profile.R | vertical profile time series
bind.vp | | bind.R
bind.vplist | | bind.R
bind.vpts | | bind.R
summary.vplist | | vplist.R
print.vplist | | vplist.R
vpts | in NAMESPACE | vpts.R | vertical profile time series
vptsHelper | | vpts.R
vintegrate | in NAMESPACE | vintegrate.R | migration traffic and vertical integration
vintegrate.vp | | vintegrate.R
vintegrate.vplist | | vintegrate.R
vintegrate.vpts | | vintegrate.R
print.vpts | | vpts.R
vol2bird | in NAMESPACE | vol2bird.R | processing vertical profiles from polar volume data
rsl2odim | in NAMESPACE | vol2bird.R | processing vertical profiles from polar volume data
rsl2odim_tempfile | | vol2bird.R
readvp.table | in NAMESPACE | vp.R | loading data
regularize | in NAMESPACE | vpts.R | vertical profile time series
mtr | in NAMESPACE | traffic.R | migration traffic and vertical integration
summary.vp | | vp.R
is.vp | in NAMESPACE | vp.R | vertical profile
is.vplist | in NAMESPACE | vplist.R | vertical profile
[...]

PR #70, opened 2018-03-08
Pkgdown reference page structure

stijnvanhoey:
This PR aims to provide a 
1. more complete, see #56  and 
2. better structured overview, see #50  of the NAMESPACE defined functions.  

I did not included the split into individual R-files of the source code itself, as I'm planning to provide that in a separate PR. This one focuses on the documentation website.


stijnvanhoey:
@adokter, good idea; adapted as requested

PR #82, opened 2018-03-27
Add MIT license

peterdesmet:
bioRad needs a software license to make it proper open source code. I opted for:

- [MIT license](https://choosealicense.com/licenses/mit/) to allow wide reuse. Is also used for [vp-processing](https://github.com/enram/vp-processing/blob/master/LICENSE), [flowviz](https://github.com/enram/bird-migration-flow-visualization/blob/master/LICENSE) or our [INBO R packages](https://github.com/inbo/wateRinfo/blob/master/LICENSE)
- `Adriaan M. Dokter and contributors` as copyright holder

The license file will be recognized by GitHub.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2018 Adriaan M. Dokter and contributors
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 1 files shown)


peterdesmet:
Just noticed that DESCRIPTION mentions:

https://github.com/adokter/bioRad/blob/6794298e3e3af90fd3d983da50407c69d1d3c242/DESCRIPTION#L15

[GPL-3](https://choosealicense.com/licenses/gpl-3.0/) is:

> a copyleft license that requires anyone who distributes your code or a derivative work to make the source available under the same terms

If we're not really concerned about that, I would opt for MIT. @adokter?

Issue #94, opened 2018-04-18
Review and update function documentation


peterdesmet:
Review for each function:

- Follows http://style.tidyverse.org/code-documentation.html styleguide. Except indentation: don't indent with two spaces for a `@param`, etc. + add empty line after each section (after all params, after export, etc.)
- Don't use word object for each object

* [ ] [beam.R](https://github.com/adokter/bioRad/blob/develop/R/beam.R)
* [ ] [bind_into_vpts.R](https://github.com/adokter/bioRad/blob/develop/R/bind_into_vpts.R)
* [ ] [bioRad-deprecated.R](https://github.com/adokter/bioRad/blob/develop/R/bioRad-deprecated.R)
* [ ] [bioRad.R](https://github.com/adokter/bioRad/blob/develop/R/bioRad.R)
* [ ] [calculate_vp.R](https://github.com/adokter/bioRad/blob/develop/R/calculate_vp.R)
* [ ] [check_night.R](https://github.com/adokter/bioRad/blob/develop/R/check_night.R)
* [ ] [color_scale.R](https://github.com/adokter/bioRad/blob/develop/R/color_scale.R)
* [ ] [composite_ppi.R](https://github.com/adokter/bioRad/blob/develop/R/composite_ppi.R)
* [ ] [datasets.R](https://github.com/adokter/bioRad/blob/develop/R/datasets.R)
* [ ] [dbz_eta.R](https://github.com/adokter/bioRad/blob/develop/R/dbz_eta.R)
* [ ] [docker.R](https://github.com/adokter/bioRad/blob/develop/R/docker.R)
* [ ] [download_basemap.R](https://github.com/adokter/bioRad/blob/develop/R/download_basemap.R)
* [ ] [download_vpfiles.R](https://github.com/adokter/bioRad/blob/develop/R/download_vpfiles.R)
* [ ] [get_angles.R](https://github.com/adokter/bioRad/blob/develop/R/get_angles.R)
* [ ] [get_quantity.R](https://github.com/adokter/bioRad/blob/develop/R/get_quantity.R)
* [ ] [get_scan.R](https://github.com/adokter/bioRad/blob/develop/R/get_scan.R)
* [ ] [integrate_profile.R](https://github.com/adokter/bioRad/blob/develop/R/integrate_profile.R)
* [ ] [map.R](https://github.com/adokter/bioRad/blob/develop/R/map.R)
[...]

Issue #84, opened 2018-03-28
Update relevant function names


peterdesmet:
Discussion on naming convention and how logically (re)name functions was done in #51. This issue is to track the work involved in actually doing this.

peterdesmet:
From `develop` I created a new branch [rename_functions](https://github.com/adokter/bioRad/tree/rename_functions), where we will work on renaming functions.

peterdesmet:
@adokter, I discovered some previously unknown (helper) functions. Can you check above (where your name is indicated) and answer questions in comment? Not that urgent.

adokter:
@peterdesmet I updated (parts) of above, off to the airport now.

Two last minute thoughts on the functions that read in data (read.pvol, read.vp ->pvol, vp). It makes sense to add in the future also functions that write these object to a hdf5 file. How would we call those functions then, if now pvol hasn't any reference to that it reads in something? So maybe `read_pvol` and `write_pvol` etc are more descriptive names after all.

peterdesmet:
Yeah, thought about that too and I would call those write functions `vpfile()` and `pvolfile()`, which is in line with creating those things.

If you prefer `read` and `write` in the names, I'd go for `pvol_read`/`pvol_write` and `vp_read`/`vp_write`.

adokter:
@peterdesmet @stijnvanhoey I know we just settled on the naming, but since it's so important and can only be done once, here still some issues I had to think of. They all relate to the (also elegant) convention that a function name returns an object with the same name. This also has disadvantages:

1) several new names of often used functions (e.g. `vp`, `vpi`,`pvol`) reveal very little about what the function actually does. I was looking at Hadley Wickham's style guide (http://r-pkgs.had.co.nz/style.html) and I like his recommendation that functions are verbs, and that the name reveals what the function does
2) it might be confusing that we now have function names and classes that have the same name. I like it in the current setup that you can pull up a help page for the function and class object separately, and in the current proposal those will be on the same page.

How about changing the new function names we agreed upon to these new new function names? Their names are more revealing about what the function does:
`vp` → `read_vp`
`vpts_from_text` → `read_vpts_from_text`
`vpi` → `integrate_altitude`
`scan` → `retrieve_scan`
`quantity` → `retrieve_quantity`
`vpts` → `bind_into_vpts`

Sorry for bringing this up at this late stage, but naming is difficult and we can only get it right once...

peterdesmet:
[...]

PR #96, opened 2018-04-18
Rename functions

peterdesmet:
This pull request contains work listed in #84. **Do not merge this PR until all boxes are checked at #84**

Process

- @stijnvanhoey moves, renames and deprecates functions
- @peterdesmet verifies if deprecation warning is shown and if all is OK, checks of tasks at #84
- @peterdesmet has also started reviewing and updating the function documentation (keeping track in #94), but the progress of that work should not stop this PR


stijnvanhoey:
As a first refactoring effort, this can be revised and - eventually - merged. Still some more work to do (cfr. issue #100, the input checks,...), but the main objective (_new function names_) according to #84 is done.

peterdesmet:
Over 250 commits and 164 files changes: what a massive PR. 😄 I reviewed and was able to check of everything in #84!  🎉There are a couple of smaller remaining issues, that I'll document, but none of these block this PR, so I'll go ahead and merge this. 🍻

adokter:
Wow cheers to that, excellent!!

Issue #118, opened 2018-04-23
duplicate alias for deprecated cmt, mt and mtr functions


adokter:
From devtools check:
```
* checking Rd metadata ... WARNING
Rd files with duplicated alias 'cmt':
  ‘bioRad-deprecated.Rd’ ‘cmt.Rd’
Rd files with duplicated alias 'mt':
  ‘bioRad-deprecated.Rd’ ‘mt.Rd’
Rd files with duplicated alias 'mtr':
  ‘bioRad-deprecated.Rd’ ‘mtr.Rd’
```

PR #120, opened 2018-04-25
Double alias fix

stijnvanhoey:
This PR fixes issue #118 by removing the duplicate `mt`, `mtr` and `mtc` function mentions. As these are actually deprecated functions, I added them to the `deprecated-bioRad.R` file


adokter:
Thanks - just fyi: in a previous commit I also copied most of the documentation of these functions to `integrate_profile.R`, such that it is also available in non-deprecated form

stijnvanhoey:
I've seen, but as this is part of deprecated, I thought it was a bit overkill to add secific templates for the `@details` section. I think having them at both locations isn't a problem here.

Issue #112, opened 2018-04-23
rename function arguments


adokter:
So far we have focussed on renaming the function names, but the function argument names also need improvements. E.g. `integrate_profile` has `alt.min` argument, which still has the dot convention we want to deprecate. This is still quite a lot of work..., but I think we cannot push anything to master until this is done.

We could contact the editors of Ecograpy to forward install instructions to the editors of the paper to a branch that has not yet renamed arguments, only function names. Then we don't have to release to the public yet and have more time.

## Functions

* [ ] beam.R: `beam_height(range, elev, k = 4/3, lat = 35, re = 6378, rp = 6357)`
* [ ] beam.R: `beam_width(range, angle = 1)`
* [ ] bioRad.R: no functions
* [ ] bioRad-deprecated: arguments should remain the same
* [ ] bind_into_vpts.R: `bind_into_vpts(x, ...)`
* [ ] bind_into_vpts.R: `bind_into_vpts(...)`
* [ ] bind_into_vpts.R: `bind_into_vpts(x, ...)`
* [ ] bind_into_vpts.R: `bind_into_vpts(..., attributes_from = 1)`
* [ ] calculate_vp.R: `calculate_vp(vol.in, vp.out = "", vol.out = "", autoconf = FALSE, verbose = FALSE, mount = dirname(vol.in), sd_vvp = 2, rcs = 11, dualpol = FALSE, rhohv = 0.95, elev.min = 0, elev.max = 90, azim.min = 0, azim.max = 360, range.min = 5000, range.max = 25000, nlayer = 20L, hlayer = 200, dealias = TRUE, nyquist.min = if (dealias) 5 else 25)`
* [ ] check_night.R: `check_night(x, ..., elev = -0.268)`
* [ ] check_night.R: `check_night(lon, lat, date, elev = -0.268)`
* [ ] check_night.R: `check_night(x, ..., elev = -0.268)`
* [ ] check_night.R: `check_night(x, ..., elev = -0.268)`
* [ ] check_night.R: `check_night(x, ..., elev = -0.268)`
* [ ] check_night.R: `check_night(x, ..., elev = -0.268)`
* [ ] composite_ppi.R: `composite_ppi(x, param = "DBZH", cells.dim = c(100, 100))`
* [ ] data.R: no functions
* [ ] dbz_eta.R: `dbz_to_eta(dbz, wavelength, Km = 0.93)`
[...]

Issue #126, opened 2018-08-01
Deprecate arguments


stijnvanhoey:
As referenced in [#112](https://github.com/adokter/bioRad/issues/51#issuecomment-374747768), we can **rename functions** while keeping the old name working as well, as follows for a function oldName(arg1_old,arg2_old that we want to replace with newName(arg1_new,arg2_new) (using roxygen):

```
#' @name oldName-deprecated
#' @rdname bioRad-deprecated
#' @section \code{oldName}:
#' For \code{oldName}, use \code{\link{newName}}.
#' @export
oldName=function(arg1_old, arg2_old){
  .Deprecated("newName")
  newName(arg1_new=arg1_old, arg2_new=arg2_old)
}
```

However, when we **only deprecate the function arguments**, without changing the function name itself, this approach does not work anymore...

I did a quick tour on possible solutions:
- https://ropensci.org/technotes/2017/01/05/package-evolution/ -> collect all arguments, and do checks
- http://r-pkgs.had.co.nz/release.html#compatibility -> argument-specific control loop on presence

I'm tending to use the latter approach. Applied to the `map.R` function `map.ppi`

```
map.ppi <- function(x, map, param, alpha = 0.7, xlim, ylim,
                    zlim = c(-20, 20), ratio, radar.size = 3,
                    radar.color = "red", n.color = 1000, ...) {
...
}
```
this becomes:

```
map.ppi <- function(x, map, quantity, param, alpha = 0.7, xlim, ylim,
                    zlim = c(-20, 20), ratio, radar_size = 3, radar.size = 3,
                    radar_color = "red", radar.color = "red", n_color = 1000, n.color = 1000, ...) {

  if (!missing(param)) {
    warning("argument param is deprecated; please use quantity instead.", call. = FALSE)
    quantity <- param
  }
  if (!missing(radar.size)) {
    warning("argument radar.size is deprecated; please use radar_size instead.", call. = FALSE)
    radar_size <- radar.size
  }
  if (!missing(radar.color)) {
    warning("argument radar.color is deprecated; please use radar_color instead.", call. = FALSE)
    radar_color <- radar.color
  }
  if (!missing(n.color)) {
    warning("argument n.color is deprecated; please use n_color instead.", call. = FALSE)
    n_color <- n.color
  }

...
}
```

[...]

PR #127, opened 2018-08-02
Argument renaming

stijnvanhoey:
This mainly handles #112 by renaming the arguments. Output of R CMD CHECK:

```
checking installed package size ... NOTE
  installed size is 34.7Mb
  sub-directories of 1Mb or more:
    data      3.1Mb
    extdata  31.2Mb
R CMD check results
0 errors | 0 warnings | 1 note 
```
There are still some elements for discussion reported on #112, but I would already start reviewing and checking the new code @peterdesmet  and @adokter


R/map.R
```diff
@@ -65,8 +68,27 @@ map <- function(x, ...) {
 #' @describeIn map plot a 'ppi' object on a map
 #' @export
 map.ppi <- function(x, map, param, alpha = 0.7, xlim, ylim,
-                    zlim = c(-20, 20), ratio, radar.size = 3,
-                    radar.color = "red", n.color = 1000, ...) {
+                    zlim = c(-20, 20), ratio, radar_size = 3, radar.size = 3,
```
peterdesmet: @stijnvanhoey arguments radar.size and n.color are still present

R/plot.vp.R
```diff
@@ -24,11 +26,25 @@
 #' @examples
 #' data(example_vp)
 #' plot(example_vp)
-#' plot(example_vp, line.col = "blue")
-plot.vp <- function(x, quantity="dens",
+#' plot(example_vp, line_col = "blue")
+plot.vp <- function(x, quantity = "dens",
                     xlab = expression("volume density [#/km"^3*"]"),
-                    ylab = "height [km]", line.col='red', line.lwd=1, ...) {
+                    ylab = "height [km]", line_col = 'red', line.col = 'red',
```
peterdesmet: line.col and line.lwd still appear as arguments

R/plot.vpi.R
```diff
@@ -46,15 +47,22 @@
 #' # plot the migration traffic rates
 #' plot(vpi)
 #' # plot the vertically integrated densities, without night shading:
-#' plot(vpi, quantity = "vid", nightshade = FALSE)
+#' plot(vpi, quantity = "vid", night_shade = FALSE)
 plot.vpi <- function(x, quantity = "mtr", xlab = "time",
                      ylab = "migration traffic rate [#/km/h]",
-                     main = "MTR", nightshade = TRUE, elev = -0.268,
-                     lat = NULL, lon = NULL, ylim = NULL, ...) {
+                     main = "MTR", night_shade = TRUE, nightshade = TRUE,
```
peterdesmet: Both night_shade and nightshade

R/plot.vpts.R
```diff
@@ -60,10 +64,37 @@
 #' # plot total reflectivity factor (rain, birds, insects together):
 #' plot(ts[1:500], ylim = c(0, 3000), quantity = "DBZH")
 plot.vpts <- function(x, xlab = "time", ylab = "height [m]", quantity = "dens",
-                      log = TRUE, barbs = TRUE, barbs.h = 10, barbs.t = 20,
-                      barbs.dens = 5, zlim, legend.ticks, main, ...) {
+                      log = TRUE, barbs = TRUE, barbs_height = 10, barbs.h = 10,
```
peterdesmet: Still double parameters

R/pvolfile.R
```diff
@@ -3,18 +3,27 @@
...
 #' otherwise FALSE
 #'
 #' @export
 #'
 #' @examples
 #' volume <- system.file("extdata", "volume.h5", package = "bioRad")
 #' is.pvolfile(volume)   #> TRUE
-is.pvolfile <- function(filename) {
-  type <- get_odim_object_type(filename)
+is.pvolfile <- function(file, filename = NULL) {
```
peterdesmet: Why is there still filename = NULL?

R/select_vpfiles.R
```diff
@@ -19,14 +19,15 @@
 #' @examples
 #' my_path <- "~/my/directory/"
 #' select_vpfiles(my_path, "2016-10-01", "2017-01-31", c("be"))
-select_vpfiles <- function(path, start_date, end_date,
-                           country = NULL, radar = NULL) {
+select_vpfiles <- function(path, date_min, date_max, country = NULL,
```
peterdesmet: Shouldn't this be directory instead of path?

R/map.R
adokter: I think this has to do with maintaining backward compatibility.  It is rather confusing though and I wonder whether we shouldn't just break backward compatibility (it's not that we have a very large user base) @peterdesmet what do you think

R/select_vpfiles.R
adokter: yes you're right, directory is better as we use that elsewhere as well

R/map.R
peterdesmet: Yeah, @stijnvanhoey told me later that it was for backward compatibility. I would drop it: including these and remembering to eventually remove them will be more confusing for us maintainers, than it being helpful to users. A wrong parameter error in old code is pretty straightforward to debug.

stijnvanhoey: see #126

R/plot.vp.R
stijnvanhoey: see #126

R/plot.vpi.R
stijnvanhoey: see #126

R/plot.vpts.R
stijnvanhoey: see #126

R/pvolfile.R
stijnvanhoey: see #126

Issue #133, opened 2018-09-13
Move and convert R/bioRad.R to vignettes/bioRad.Rmd


peterdesmet:
[R/bioRad.R](https://github.com/adokter/bioRad/blob/develop/R/bioRad.R) mainly provides an overview of bioRad functionality. I would propose to:

1. Remove the file or if it is required, keep it minimal, cf. https://github.com/tidyverse/ggplot2/blob/master/R/ggplot2.r
2. Move the overview it currently provides to a new file `vignettes/bioRad.Rmd`

That way:

1. pkgdown will automatically detect `vignettes/bioRad.Rmd` and add it as a "getting started" page in the website.
2. It will be easier to maintain, as most of the content is text. Note that it seems that links to referenced functions will still be automatically created as long as they are in backticks, see for example: http://pkgdown.r-lib.org/articles/pkgdown.html
3. We don't have to list `bioRad.R` in the functions list.

@adokter can you let me know what you think? If you agree, let me know if a `bioRad.R` file is required and I can try to do the move.

adokter:
@peterdesmet I think a `bioRad.R` file is mandatory, it's used only for documenting the package, as a place to start for users. See also https://cran.r-project.org/web/packages/roxygen2/vignettes/rd.html under heading packages, this is the format I followed. I think it's useful to give an overview of the package functionality in this file (in addition to the vignettes). This help page for the package now shows if you do `?bioRad` which is a useful feature I think.

But see also http://r-pkgs.had.co.nz/man.html#man-packages

Maybe try adding
 ```
#' @doctype package
#' @name bioRad
```
Those flags are not present now. Maybe they will suppress it ending up in the function list as well, and point roxygen in the right direction.

**Comment by adokter** [comment]
But if above does not work, not against shortening it or making it a vignette. People should just be pointed to an overview if they do `?bioRad`, I don't have strong opinions about how exactly

**Comment by peterdesmet** [comment]
From the [link you referenced](https://cran.r-project.org/web/packages/roxygen2/vignettes/rd.html) it seems a `bioRad.R` file is not mandatory, but useful. It looks like the minimal form is:

```R
#' @keywords internal
"_PACKAGE"
```

I just tried this minimal version and did a `?bioRad`, which results in:

<img width="751" alt="screen shot 2018-09-14 at 10 49 18" src="https://user-images.githubusercontent.com/600993/45540857-cab8e280-b80d-11e8-8860-cb251caa75fe.png">

[...]

Issue #132, opened 2018-09-13
Create get_param() function


peterdesmet:
I might be missing something, but is there an intuitive way to get a `param` from a `scan`? It seems possible with e.g.:

```r
example_scan$params[[1]]
```

But it might be good to have a:

```r
get_param(scan, param = "DBZH")
```

It will probably be similar in functionality as `get_quantity()`

adokter:
closed by commit 243129220741df659844a49962b76b2e7e66be6c

peterdesmet:
Note: new public functions should be explicitly added to reference overview in _pkgdown.yaml. I will do this. Reopening.

peterdesmet:
Closed in 3ac2e68

Issue #113, opened 2018-04-23
reduce package size


adokter:
Total package on CRAN should be less than 5Mb

This devtools check NOTE should be avoided, suggesting no data folder > 1Mb
```
checking installed package size ... NOTE
  installed size is 34.6Mb
  sub-directories of 1Mb or more:
    data      3.1Mb
    extdata  31.2Mb
```

adokter:
We might have to consider a separate data package to come with bioRad, as even a single radar file is much larger than 1 Mb. See https://journal.r-project.org/archive/2017/RJ-2017-026/index.html

peterdesmet:
If you let me know which data should be included in the package, I can easily create one on Zenodo.

peterdesmet:
@adokter agree. Having everything in this repository simplifies things. Hope you can reduce the large files while still keeping them representative and useful for examples.

adokter:
closed by commits 72d34d83feaf6078822f23c3aecd657814bde79d, 6e12aada7be4d20452dddb6ac3c5047a79f290e0 and 0c20d6f27651a0060a9db6ff5870805eca059895

adokter:
documentation changes made the package exceed 5 Mb again
```

**Comment by adokter** [comment]
* checking installed package size ... NOTE
  installed size is  5.1Mb
  sub-directories of 1Mb or more:
    data      1.6Mb
    extdata   2.6Mb
```

adokter:
Let's try to get it approved on CRAN first before putting further effort into reducing the package size, closing for now

PR #147, opened 2018-09-25
Develop

peterdesmet:
- Reduce data size (#113)
- List get_param (#132)
- Reference new paper
- Use default .gitignore
- Change role of @peterdesmet @stijnvanhoey 
- Reduce bioRad.R to essentials (#133)


peterdesmet:
@adokter can you have a look and merge if ok?

PR #151, opened 2018-09-27
Add biocViews to automatically look for rhdf5

peterdesmet:
From [this stack exchange](https://bioinformatics.stackexchange.com/questions/3365/r-package-development-how-does-one-automatically-install-bioconductor-packages) it looks like including `biocViews:` to the DESCRIPTION allows the installation to look for bioconductor dependencies as well. If true, it means users don't have to do:

```r
source("http://bioconductor.org/biocLite.R")
biocLite("rhdf5")
````

before installing bioRad, which would simplify things nicely. I'll test this first.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: bioRad
 Title: Biological analysis and visualization of weather radar data
-Version: 0.3.1
+Version: 0.3.0.9151
 Description: Extracts and visualizes biological signals from weather radar data.
 Authors@R: c(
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by peterdesmet** [comment]
Just removed `rhdf5` and `bioRad` from my computer and used:

```R
devtools::install_github("adokter/bioRad", ref = "biocviews")
```

And the installation worked without problem! :clap: :clap: One more instruction we can remove! I'll bump dev version and merge this to master.

Issue #158, opened 2018-10-01
Figure out <details> syntax for pkgdown


peterdesmet:
The details syntax works in Markdown, but not in the rendered website. Check if this can be fixed.

peterdesmet:
Duplicate of #300.

PR #163, opened 2018-10-22
#158 Source switch

BerendWijers:
Hi,

I've added a check to see if it needs to pass a bounding box or center location to get_map.

Open street maps has some issues with invalid tokens, so this source wont work for me.
Stamen maps does work as intended.
Google I cant test as I do not have an api_key. Anyways, the google api_key should be able to be passed with the "..." to get_map and all should be fine.

I've done the devtools::check()
`Updating bioRad documentation
Loading bioRad
Welcome to bioRad version 0.3.0.9160
Docker daemon running, Docker functionality enabled.
Error in loadNamespace(name) : there is no package called ‘testthat’`

Also, I dont think any documentation needs to be changed as this time source is used within the download_basemap instead of directly being passed to get_map. For the user, this changes nothing (I think).

Best regards,
Berend-Christiaan Wijers

### DIFF
--- a/R/download_basemap.R
+++ b/R/download_basemap.R
@@ -9,6 +9,7 @@
 #'   ppi extent is selected automatically.
 #' @param alpha Transparency of the basemap (0-1).
 #' @param verbose Logical, whether to print information to console.
+#' @param source "google","osm", "stamen". If source != google, adapt location to provide boundingbox instead of center.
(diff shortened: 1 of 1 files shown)


adokter:
Thanks Berend-Christiaan!

merged by commit 5bdc6c67f9c0ddb672fb1c904587c7c101e37674

Issue #172, opened 2018-12-04
Don't unzip files with download_vpfiles()


peterdesmet:
`download_vpfiles()` currently not only downloads (monthly zips of) vp files, but **unzips these as well**. That makes the function slower than just downloading (see Performance below). I would suggest to remove:

https://github.com/adokter/bioRad/blob/e9355d7bc4f6df0d4b5db2a950fd46854f62d87b/R/download_vpfiles.R#L103-L105

To speed up the download function, and let users unzip the files themselves afterwards (either using R or using their OS). @stijnvanhoey @adokter @cecilianilsson709 any objections?

## Directory structure

The reason this was implemented as such is so that the local directory structure would mimic that of the ENRAM data repository:

```
bejab201808.zip
bejab201809.zip
bewid201808.zip
bewid201809.zip
be/
    jab/
        2018/
            08/
            09/
    wid/
        2018/
            08/
            09/
```

Rather than the default unzipping to:

```
bejab201808.zip
bejab201809.zip
bewid201808.zip
bewid201809.zip
bejab201808/
bejab201809/
bewid201808/
bewid201809/
```

That structure however is not required for `select_vpfiles()`: it searches the data directory recursively based on file names, independent from any directory structure:

```R
length(select_vpfiles(
  directory = "my_data/download",
  date_min = "2018-08-01",
  date_max = "2018-08-31",
  country = c("be")
))
# 17522

# After unzipping zips
length(select_vpfiles(
  directory = "my_data/download_and_manually_unzip",
  date_min = "2018-08-01",
  date_max = "2018-08-31",
  country = c("be")
))
# 17522
```

### With zipping

```R
library(bioRad)
start_time <- Sys.time()
download_vpfiles(
  "2018-08-01",
  "2018-09-30",
  c("be"),
  c("jab", "wid"),
  directory = "my_data/download"
)
end_time <- Sys.time()
difftime(end_time, start_time, units = "secs")
```

Time difference of 355.2779 secs

### Without zipping

```
library(bioRad)
start_time <- Sys.time()
download_vpfiles(
  "2018-08-01",
  "2018-09-30",
  c("be"),
  c("jab", "wid"),
  directory = "my_data/download_and_manually_unzip"
)
end_time <- Sys.time()
difftime(end_time, start_time, units = "secs")
```

Time difference of 202.2632 secs

So a (substantial) time difference of 153 seconds in the case of 17522 files.

adokter:
[...]

PR #176, opened 2018-12-08
Update download_vpfiles() and select_vpfiles()

### LINKED ISSUE #131: check whether we can reduce number of dependencies
We currently depend on 14 packages, which is quite a lot.

Check whether we can reduce this number
https://github.com/adokter/bioRad/blob/master/DESCRIPTION#L18-L31

### LINKED ISSUE #150: Use httr instead of (r)curl
From the [rOpenSci onboarding guide](https://ropensci.github.io/dev_guide/building.html#recommended-scaffolding):

> For http requests we strongly recommend using httr over RCurl.

We need to tackle RCurl vs curl for #131 anyway, so might be good to use httr instead.

### LINKED ISSUE #153: select_vpfiles: add data dir to file paths
```R
download_vpfiles("2016-10-01", "2016-10-07", "fr", c("tro"))
unzip("frtro201610.zip", exdir = "frtro201610")
vpfiles <- select_vpfiles("frtro201610", "2016-10-01", "2016-10-07")
read_vpfiles(vpfiles)
```

Will generate an empty list, because select_vpfiles does not prepend the file paths with the directory to the file: `01/06/frtro_vp_20161001T0600Z.h5` instead of `frtro201610/01/06/frtro_vp_20161001T0600Z.h5`, so the user needs to prepend this themselves: https://enram.github.io/vp-processing/extract-vp.html#select-date-range-radars

I think it would be more user-friendly if the full path is included...

### LINKED ISSUE #173: Use 5 letter radar codes and drop "country" argument in download_vpfiles() and select_vpfiles()
`download_vpfiles()` currently requires both the argument `radar` and `country` to be provided. But it combines them in a way that leads to non-existing radars and thus files that cannot be downloaded:

```R
download_vpfiles(
  date_min = "2016-10-01",
  date_max = "2016-12-31",
  country = c("be", "se"),
  radar = c("jab", "wid", "ang"),
  directory = "my_data"
)
```

Will search for:

```
bejab: ok
sejab: error
bewid: ok
sewid: error
beang: error
seang: ok
```

Given that **both arguments are required anyway** (radar without country does not work and vice versa) and to avoid against creating radar codes that do not exist, I would argue to drop the `country` parameter, and let the user explicitly provide 5 letter radar codes in `radar`:

```R
download_vpfiles(
  date_min = "2016-10-01",
  date_max = "2016-12-31",
  radar = c("bejab", "bewid", "seang"),
  directory = "my_data"
)
```

---

For the parallel function `select_vpfiles()`

```R
select_vpfiles(
  date_min = "2016-10-01",
  date_max = "2016-12-31",
#  country = c("be"),
  radar = c("wid"),
  directory = "my_data"
)
```

The radar and country ARE optional, i.e. one can correctly search for dates, dates + radars, dates + countries or all 3. Still, I would opt here too to drop the country parameter and ask for 5 letter radar codes, because:

- 3 letter radar codes _can_ be shared between countries, leading to unexpected results. Better if the user is explicit
- I think most analyses want to be explicit about the radars they select
- That way we can keep `download_vpfiles()` and `select_vpfiles()` similar (e.g. you can reuse a radar code list to download and select)

I don't think this will break current functionality. @adokter @stijnvanhoey @CeciliaNilsson709 comments?

peterdesmet:
## For both functions

- Drop country argument and use 5 letter radar codes. Parameter `radar` renamed to `radars`. Fixes #173 👌
- Raise error if radar code is not 5 letters
- Gets rid of RCurl and lubridate. Fixes #131 👌
- Update function documentation
- Update function deprecation

## download_vpfiles()

Updated syntax:

```R
download_vpfiles(
  date_min = "2016-10-03", # required
  date_max = "2016-10-05", # required
  radars = c("bejab", "bewid"), # required
  directory = "my_data" # optional, default: "."
  overwrite = TRUE # optional, default: FALSE
)
```

- Unzip to 5 letter radar code directory. Fixes https://github.com/adokter/bioRad/issues/172#issuecomment-444264766 👌
- Add new optional parameter `overwrite = FALSE` (default) to skip downloading previously downloaded files 🎉
- Just use curl, not httr. Fixes #150 👌
- Simplify path, url, file name creation in function
- Nicer messaging of download status, using `message()` not `print()` so they can be ignored in Rmd 🎉

    ```
    bewid201710.zip: already downloaded
    bejab201711.zip: http error 404
    bewid201711.zip: successfully downloaded
    ```

## select_vpfiles()

Updated syntax:

```R
select_vpfiles(
  date_min = "2016-10-03", # optional
  date_max = "2016-10-05", # optional
  radars = c("bejab", "bewid"), # optional
  directory = "my_data" # optional, default "." and moved to last position to align with download_vpfiles()
)
```

- Prepend directory to returned paths, making these directly readable with `read_vpfiles()`. Fixes #153 👌
- Make `date_min` and `date_max` optional parameters (will only be used if both are set)


R/download_vpfiles.R
```diff
@@ -1,109 +1,99 @@
...
-  # create date range set of potential downloadable zip files (if all data
-  # would exist)
-  start <- floor_date(as_date(date_min, tz = NULL), "month")
-  end <- floor_date(as_date(date_max, tz = NULL), "month")
-  dates_to_check <- seq(start, end, by = "months")
+download_vpfiles <- function(date_min, date_max, radars, directory = ".",
+                             overwrite = FALSE) {
+  # Stop if radar codes don't contain exactly 5 characters
+  wrong_codes <- radars[nchar(radars) != 5]
+  if (length(wrong_codes) > 0) {
```
stijnvanhoey: Do we have an overview list of the current radars and countries (e.g. compiled from the [overview list](https://lw-enram.s3-eu-west-1.amazonaws.com/radars.csv) to check the existence of radars and countries? Currently, every 5-string combination will be attempted for download.

R/select_vpfiles.R
```diff
@@ -1,66 +1,85 @@
...
-  if (is.null(country)) {
-    country <- "([a-z]{2})"
+#' my_path <- "~/my/directory"
+#' select_vpfiles(
+#'   date_min = "2016-10-03",
+#'   date_max = "2016-10-05",
+#'   radars = "bejab",
+#'   directory = my_path
+#' )
+select_vpfiles <- function(date_min = NULL, date_max = NULL, radars = NULL,
```
stijnvanhoey: When people do not ass the keyword arguments, sequence matters in R. So make sure that the code is backward compatible instead of switching argument position of funcitons in between releases

R/select_vpfiles.R
```diff
@@ -1,66 +1,85 @@
...
+                           directory = ".") {
+  # Stop if radar codes don't contain exactly 5 characters
+  wrong_codes <- radars[nchar(radars) != 5]
```
stijnvanhoey: Duplicate code from above. Consider to create a small helper function (not exported to namespace) within this file itself to check incoming `radars` arguments.

R/select_vpfiles.R
```diff
@@ -1,66 +1,85 @@
...
+  if (length(wrong_codes) > 0) {
+    stop("Radar codes should contain exactly 5 letters: ",
+         paste(wrong_codes, collapse = ", "))
   }
-  if (is.null(radar)) {
-    radar <- "([a-z]{3})"
+
+  # If radars not defined, create regex for any 5 letters lowercase code
```
stijnvanhoey: why?

R/select_vpfiles.R
```diff
@@ -1,66 +1,85 @@
...
-    country <- "([a-z]{2})"
+#' my_path <- "~/my/directory"
+#' select_vpfiles(
+#'   date_min = "2016-10-03",
+#'   date_max = "2016-10-05",
+#'   radars = "bejab",
+#'   directory = my_path
+#' )
+select_vpfiles <- function(date_min = NULL, date_max = NULL, radars = NULL,
+                           directory = ".") {
```
stijnvanhoey: add unit tests for both ingle name radar as input `select_vpfiles("bejab")` as well as vector `select_vpfiles(c("bejab", "fikor"))`

stijnvanhoey: What with backward compatibility here? Could we do this similarly to previous efforts and make sure people using `radar` and `country` are either handled as well or getting a proper depcrecation warning?

R/select_vpfiles.R
stijnvanhoey: A bit confused by https://github.com/adokter/bioRad/issues/173#issuecomment-445461125 but maybe it' only a Python thing that named arguments should not preceed not-named arguments...

R/select_vpfiles.R
peterdesmet: Yeah, thought the same. Will do, which of the two files do you suggest?

R/select_vpfiles.R
peterdesmet: We basically want to answer:

1. Do we want to align the order of `select_vpfiles()` and `download_vpfiles()`. My 2 cents: yes.
2. If so, only the position of `directory` (and parameter `overwrite`) is different: do we want it in front or back. My 2 cents: choose order `date_min`, `date_max`, `radars`, `directory`, (`overwrite`) because https://github.com/adokter/bioRad/issues/173#issuecomment-445495121).
3. And for backwards compatibility: several things have changed: `radar` -> `radars`, `country` -> included in `radars`, `directory` -> to end. All these are now [captured in the deprecation](https://github.com/adokter/bioRad/pull/176/files#diff-2a6c884158d6474f545d10290280f1bb) of `retrieve_vp_paths()` and `download_vp()`. Since this change follows fairly quickly after our latest deprecation and any error message is pretty straightforward, I though to bundle these with our latest deprecation.

R/download_vpfiles.R
peterdesmet: > Currently, every 5-string combination will be attempted for download.

Every 5-letter combination **that the user provides** will be attempted for download, so they are in control. I agree it would be nice to know which radars exist, but so does information about their coverage (for which you also get 404s).

I would suggest to bundle this at some point in a separate function `coverage_vpfiles()` that could provide such information for remote or even downloaded vp files. @CeciliaNilsson709 has already created some nice charts to express the coverage. 👌 And the returned information (list of vp files that are available) could optionally be passed to the `download_vpfiles()` for more targeted download. So something for a separate enhancement issue I would say.

peterdesmet:
> As we will have to do it anyway, please provide unit tests directly when refactoring/adding funcitons. If you need any help writing the tests, just let me know.

I agree completely, but have never done this so I would need a crash course from you. 😁

R/download_vpfiles.R
```diff
@@ -1,109 +1,99 @@
...
-      print(paste("Downloading file", countryradardate[i]))
-      curl_download(urls[i], file.path(directory, countryradardate[i]),
-        quiet = FALSE
-      )
+  # Start download and unzipping
+  message(paste("Downloading data from", base_url))
+
+  for (radar_year_month in radar_year_months) {
+    # Create filename of format bejab201610.zip (removing _ and /)
+    file_name <- paste0(gsub("/", "", gsub("_", "", radar_year_month)), ".zip")
```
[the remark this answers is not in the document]
peterdesmet: I needed a different separator than `/`, because the `year/month` is pasted to it as well. Sometimes I need `radar/year/month` (just replacing `_`).

R/download_vpfiles.R
```diff
@@ -1,109 +1,99 @@
...
-  collapse = "/"
+  # Split 5 letter radar codes into format be_jab
+  ra_dars <- paste(substring(radars, 1, 2), substring(radars, 3, 5),
+                       sep = "_")
+
+  # Create series of yyyy/mm based on date_min/max: 2016/10, 2016/11, 2016/12
+  dates <- seq(
+    as.Date(date_min, tz = NULL),
+    as.Date(date_max, tz = NULL),
+    by = "months"
```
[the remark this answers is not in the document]
peterdesmet: Good catch, it doesn't. Will fix this. Agree we need unit tests for this asap.

R/select_vpfiles.R
peterdesmet: To match files of the format `[a-z]{5}_vp_yyyymmdd*.*`

R/download_vpfiles.R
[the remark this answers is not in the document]
peterdesmet: If we don't want to use `floor_date()` from lubridate, how do you feel about string manipulation:

```R
> date_min <- "2016-09-25"
> as.Date(paste(substring(date_min, 1, 7), "01", sep = "-"), tz = NULL)
[1] "2016-09-01"
```

adokter:
> I agree completely, but have never done this so I would need a crash course from you. 😁

Same here!

R/select_vpfiles.R
stijnvanhoey: I would propose the download version (first one someone would encounter), but I do not have strong opinion

peterdesmet: Done in https://github.com/adokter/bioRad/pull/176/commits/a696438d213e58d06e009f3a7cbb0d438bec8f82

R/select_vpfiles.R
peterdesmet: Moved testing to other issue so I can close here: https://github.com/adokter/bioRad/issues/182

R/download_vpfiles.R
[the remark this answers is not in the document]
peterdesmet: Helper function to check in 8c9410e. Now also covers #63 again.

stijnvanhoey:
Nice work @peterdesmet

Issue #224, opened 2019-04-23
Dev vs stable documentation


peterdesmet:
> @adokter:
> 
> Today I committed this PR (#223), bumping bioRad to version 0.4.0.9223. It adds several functions and updates the documentation. Do we want to regenerate the roxygen site also for these development releases?
> 
> It seems in an ideal world you would want to keep to versions of the site, one for the official CRAN release, and one for the latest development release.
> 
> But if we have to choose for now, what is the best practice? I'm inclined to keep the site up to date with the latest release, even when it's a development release.

I agree to also generate for dev versions, but will look into having documentation for both stable and latest versions. 👍

peterdesmet:
pkgdown supports the creation of development vs stable documentation by setting `development: mode: auto` in `_pkgdown.yml`.

It results in `http://adokter.github.io/bioRad` showing the documentation of the latest **stable** version, i.e. any `x.x.x` release:

![Biological Analysis and Visualization of Weather Radar Data • bioRad (1)](https://user-images.githubusercontent.com/600993/56638992-defc7200-666f-11e9-8b44-9fe07006a904.png)

While documentation for a dev version `x.x.x.9xxx` will be rebuilt at `http://adokter.github.io/bioRad/dev`:

![Biological Analysis and Visualization of Weather Radar Data • bioRad (2)](https://user-images.githubusercontent.com/600993/56639035-fb98aa00-666f-11e9-9ad9-e8a03680774f.png)

So, not necessarily a difference between a minor (`x.x.x`) and a CRAN release, but close enough. As long as we release (minor) versions on CRAN and keep the rest as dev versions, then we have a solution in which `http://adokter.github.io/bioRad` shows the CRAN version and `http://adokter.github.io/bioRad/dev` the latest version.

I like that approach. @adokter what do you think?

peterdesmet:
I created a PR that implements this: #225

adokter:
Nice solution 👍

peterdesmet:
Now live:

http://adriaandokter.com/bioRad/
http://adriaandokter.com/bioRad/dev/

adokter:
Nice! We probably should add crosslinks to these pages, so people now how to find them

peterdesmet:
Indeed, can be done in README.Rmd

peterdesmet:
[...]

PR #226, opened 2019-04-24
Add development site

adokter:
See issue #224 for details


peterdesmet:
Ok for me to merge this to master, so we can see the effect.

adokter:
> Ok for me to merge this to master, so we can see the effect.

Ok will do that in a minute

New function `vol2bird_version()` should be under *Creating vertical profiles of biological targets*

We should also add a message on the main site that there is now a dev site as well

Issue #233, opened 2019-06-12
change definition of height-integrated ground speed ff


adokter:
See issue #232 for rationale.

Previous definition was `ff = sqrt(u^2 + v^2)`, with `u`,`v` the height-averaged speed components.

In the new definition, we take a density-weighted average of layer-specific speed values `ff`.

These two definitions are different as soon as the direction of movement varies with altitude.

adokter:
closed by commit https://github.com/adokter/bioRad/commit/72be6d1d1b8b2e5a571c593c7cfcb492e71f8218

Issue #232, opened 2019-06-11
clarify definition height-integrated velocity and traffic rates


adokter:
There are two ways to calculate a height-average of speeds:

1) calculating a density weighted-average of the height layer speeds `ff`
2) calculating density weighted-averages of the height layer speed components `u` and `v`, and then taking `sqrt(u^2+v^2)` for the speed

These two are different, because (1) refers to the average ground speed, while (2) refers to the net speed at which all biomass in the air column displaces itself relative to the ground. These two diverge when the direction of movement varies with altitude.

Currently `ff` in the result of [`integrate_profile()`](https://github.com/adokter/bioRad/blob/master/R/integrate_profile.R) refers to definition (2). We might need to change to definition (1), because it is more biologically intuitive, and because when [`alpha=NA`](https://github.com/adokter/bioRad/blob/966a1c28dda4fec02fb51cff0041017095c40d7f/R/integrate_profile.R#L11) the intuitive equalities `mtr = ff*vid` and `rtr = ff*vir` in vertically integrated profiles hold for definition (1) but not for definition (2).

On the other hand, height-integrated direction `dd` is calculated from height-integrated `u` and `v`, and therefore more closely aligned with definition (2).

> A solution would be to include two speed values, one for each speed definition. A suitable candidate for such a second speed value would be the density weighted-average of `gg=ff*cos(dd-alpha)`, i.e. the speed component perpendicular to the transect, as for that quantity `mtr = gg*vid` and `rtr = gg*vir` would hold for any value of transect direction `alpha`

Decided against above suggestion, because it only adds redundant information that is fully contained in height-averaged `u` and `v`, see below

adokter:
This issue is relevant for time-integration implementation (#77), specifically the order of the time and height integrations.

adokter:
Note that if `dd=atan2(u,v)` then `ff*cos(dd-alpha)` can be written as `u*cos(alpha)+v*sin(alpha)`, see also use of [cosfactor](https://github.com/adokter/bioRad/blob/966a1c28dda4fec02fb51cff0041017095c40d7f/R/integrate_profile.R#L155) in `integrate_profile()` implementation.

adokter:
we can calculate reflectivity traffic `rt` from `rtr` by a time-integration: 

[...]

Issue #208, opened 2019-03-26
add function to calculate Nyquist velocity from PRF


adokter:
* single PRF: `V=wavelength * PRF/4`
* dual PRF, with low and high unambiguous velocity `Vl`, `Vh`: `(Vh * Vl)/(Vh - Vl)`

adokter:
https://github.com/adokter/bioRad/commit/4ff11e8d8d7570adce22e387e68f30ea2934469c

PR #238, opened 2019-07-22
range bias correction

adokter:
This PR adds range-correction functionality and speed improvements for projecting data as PPIs

* new function `integrate_to_ppi()` to calculate a range-corrected `ppi` from a `pvol` and its associated `vp`.
* new function `scan_to_raster()` to convert a `scan` into a `RasterBrick` compatible with package raster
* new function `scan_to_spatial()` to convert a `scan` into a `SpatialPointsDataFrame` compatible with package sp
* new function `nyquist_velocity()` to calculate the unambiguous velocity of Doppler radar from its pulse repetition frequency/frequencies (#208)
* new functions `beam_distance()`, `beam_range()` to relate range (i.e. slant range), distance (i.e. down range) and height of the radar beam.
* new function `beam_profile()` to calculate for a set of beam elevations the altitudinal normalized distribution of radiated energy by those beams.
* new function `beam_profile_overlap()` to calculate the distribution overlap (in terms of Bhattacharyya distance)  between a vertical profile (`vp`) and the vertical radiation profile of a set of emitted radar beams (given by `beam_profile()`)
* `project_as_ppi()` function is now much faster (e420e5d)
* `project_as_ppi()` now accounts for earth's curvature (820e85f)
* corrected the definition of height-integrated velocity (issues #232, #233, 72be6d1)
* improved documentation of how mtr can be calculated from vid, u, v (6dce625)
* `read_pvolfile()` now also reads quantities DBZ, TH, T, because these often occur in European data and are relevant for biological analysis (note that DBZ and T are not ODIM-complient names) (5db08bd)

@peterdesmet I will merge this to master before the end of the week, because I want to use (some) of the new functionality and speed improvements during next week's radar training school in Oklahoma. Not expecting your full review of this, but just to let you know. Documentation of the new functions isn't ideal yet, but feel free to make edits already.


R/beam.R
```diff
@@ -24,26 +25,235 @@
...
+#' @param antenna numeric. Height of the centre of the radar antenna in meters
+#' @param height numeric. Height in meter.
+#'
+#' @return numeric.
+#'
+#' @details Beam profile is calculated using \link{beam_height} and \link{beam_width}. \code{height} and
+#' \code{antenna} should be given in reference to the same reference plane (e.g. ground level or sea level)
+#'
+#' @keywords internal
+gaussian_beam_profile=function(height,range,elev,antenna=0,beam_angle=1, k=4/3, lat=35, re = 6378, rp = 6357){
```
peterdesmet: @adokter to check if you're using consistent assignment (`<-`) and spacing, you can run:

```r
library(styler)
style_pkg()
```

after your edits.

R/project_as_corrected_ppi.R
```diff
@@ -0,0 +1,203 @@
...
+#' plot(my_ppi,param="vid",zlim=c(0,200))
+#' #' # download a basemap, and map the ppi:
+#' bm = download_basemap(my_ppi)
+#' map(my_ppi, bm)
+#' # calculate the range-bias corrected ppi on an even finer 500m x 500m pixel raster,
+#' # cropping the area up to 50000 meter from the radar.
+#' my_ppi <- project_as_corrected_ppi(example_pvol,example_vp,res=500,
+#' xlim=c(-50000,50000),ylim=c(-50000,50000))
+#' plot(my_ppi,param="vid",zlim=c(0,200))
+project_as_corrected_ppi = function(pvol,vp,nx=100,ny=100,xlim,ylim, zlim=c(0,4000), res, param="DBZH", lat, lon, antenna, beam_angle=1,crs,quantity=c("vir","vid","correction_factor","overlap"), k=4/3, re = 6378, rp = 6357){
```
peterdesmet: @adokter `project_as_corrected_ppi` and `project_as_ppi` sound so similar to me that I though this could be simplified by adding a "as corrected" parameter to `project_as_ppi`. But it turns out `project_as_corrected_ppi` uses a `pvol` as data input, while `project_as_ppi` only works for `scan` and `param`. Not really sure what to suggest to improve this, but the names are a bit confusing.

peterdesmet:
Extra notes:

- The vignette is not referenced in the pkgdown website (maybe on purpose)
- Would create an issue to not forget checking the documentation
- Fine to merge this into master so you can use it for the training school 👍

R/beam.R
peterdesmet: I've now run styler

adokter: thanks!

R/project_as_corrected_ppi.R
adokter: @peterdesmet I see your point. The output is still a ppi object, which is why I used the `project_as_..._ppi`  name, but definitely a lot more is happening in `project_as_corrected_ppi` than in `project_as_ppi` - it is effectively a vertical integration over all altitudes while accounting for the varying overlap of the radar beam with distance from the radar, not a simple projection.

How about `calculate_vp_corrected_ppi`?

peterdesmet: Thanks for explaining! The `calculate_vp_etc` name is probably too similar to the very different function `calculate_vp`. I suggest `integrate_to_ppi`:

- Clear that it is more than just projection
- Clear that output is ppi
- Clear that providing a single scan won't do: whole pvol is needed
- Clear that `project_to_ppi` is more basic (and just used for single scans)
- Has similarities to `integrate_profile` which also height integrates

adokter: I like it, let's go with that!

adokter:
* added the vignette to the pkgdown website (d08578e)
* added new functions to the pkgdown website function list (d08578e)
* bumped version to 0.4.0.9238, matching this PR #238
* renamed `project_as_corrected_ppi()` to `integrate_to_ppi()`
* opened issue #239 to not forget finishing the documentation

peterdesmet:
Looks all good. I've update the base branch for this PR to `develop`. I'll then merge that into master.

peterdesmet:
Vignette is now available at http://adriaandokter.com/bioRad/dev/articles/range_correction.html

PR #240, opened 2019-07-24
Create site for 0.4.0.9238

peterdesmet:
@adokter, i cannot build the `range_correction` vignette:

```
Reading 'vignettes/range_correction.Rmd'
Error in calculate_vp(pvolfile) : 
  failed to start vol2bird Docker container
```

My docker is running. Any idea what might be wrong?


adokter:
> My docker is running. Any idea what might be wrong?

Don't know - it works on my machine. Can you try to run the example in the documentation of `calculate_vp()` and see whether that gives any clues?

adokter:
You can also remove the line with `calculate_vp()` from the vignette, or put it into comments, as it is not essential

peterdesmet:
Seems the path was not available to docker:

```
docker: Error response from daemon: Mounts denied: 
The path /Library/Frameworks/R.framework/Versions/3.5/Resources/library/bioRad/extdata
```

But I think it makes more sense to comment out that part anyway, so I'll do that.

PR #241, opened 2019-07-26
Add datetime selection for vpts

adokter:
Adding a function to filter vpts objects for date. In practicals it keeps coming up that selecting only parts of a vpts object is not user-friendly, so adding some functionality.

Needs to be merged before Monday July 29 (Oklahoma training school coming up)


adokter:
@peterdesmet added `filter_datetime()` to make date/time selections in vpts objects, please check whether it matches your ideas of data filtering (cf. https://github.com/enram/vp-processing)

peterdesmet:
Won’t be able to review until Tuesday, so merge away. ☺️

adokter:
I'm renaming `filter_datetime` to `filter_vpts`. While this function currently only operates on the datetime field, it would be easy to add a `quantity` keyword, and make the function filter also on other quantities, such as height and flight direction.

peterdesmet:
Good function name. Won't be able to review further.

PR #259, opened 2019-09-03
Beam profile patch

### LINKED ISSUE #258: beam profile provides incorrect results
This plot shows the beam profile function has a strange pattern, this is due to the `range` for each elevation angle being used for each call to `gaussian_beam_profile`
``` r
require(bioRad)
#> Welcome to bioRad version 0.4.0.9243
#> Warning: Docker daemon not found
#> Warning: bioRad functionality requiring Docker has been disabled
#> 
#> To enable Docker functionality, first install Docker
plot(beam_profile(1:3000, 35000, c(1:4,25)),1:3000)
```

<sup>Created on 2019-09-03 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

bart1:
This push request should resolve #258 ( @BerendWijers have a look if this works as expected for you )


adokter:
Hi @bart1, had to revert this one for now, it breaks the default example of `integrate_to_ppi()`. But definitely a bug in `beam_profile()` that needs to be fixed

PR #260, opened 2019-09-03
Revert "Beam profile patch"

adokter:
Reverts adokter/bioRad#259


adokter:
PR broke integrate_to_ppi() function, reverting ...

Issue #251, opened 2019-08-27
make eta_observed available in range bias correction output


adokter:
Following discussions at UvA on August 27 2019, change output field of `integrate_to_ppi()`:
* rename `correction_factor` to something that avoids the word 'correction'
* add `eta_sum_expected` and `eta_sum_observed` as output field

adokter:
suggestion Bart:
correction = (VID / sum_eta_exp) * sum_eta_obs = (donut shape) * sum_eta_obs
To see effect of the correction, we compare sum_eta_obs and correction

bart1:
The availability was already largely solved I just included the last few changes in #257

adokter:
Thanks @bart1, updated defaults and documentation in commit 4123ccc8d636d1b7cff9469d8d3663e62a02a632, closing

Issue #248, opened 2019-08-07
accept ODIM files with missing /how group


adokter:
Missing /how group means the radar wavelength is not available as an attribute, however since encountering files without this group is relatively common, it is better to not make it a hard requirement for reading ODIM files

adokter:
closed by https://github.com/adokter/bioRad/commit/7729060339796576af1f14e6bdd9ce5041aeb581

PR #261, opened 2019-09-03
hotfix beam_profile() function, more output fields integrate_to_ppi()

adokter:
* hotfix of function `beam_profile()`, which contained a bug in the range calculation of different beam elevations. Bug only showed up numerically when higher beam elevations were included (#258).
* `read_pvolfile()` now accepts odim hdf5 files with a missing /how group (#248)
* bugfix of warnings in `plot.scan()` that emerged with new R release
* `integrate_to_ppi()` now includes `eta_sum` and `eta_sum_expected` (#251)


adokter:
@BerendWijers @bart1 this PR is now included in the latest development version on the master branch (0.4.0.9261), and includes fixes for issues #251 and #258

bart1:
@adokter thanks for fixing we notice something was wrong while working with the code after creating the patch but were not able to fix it yet

PR #266, opened 2019-09-15
Grid input

bart1:
The following changes should allow to use a raster of any projection to be specified for intergrate_to_ppi and scan_to_raster. The test check if the same output is calculated as when no raster is specified. It must be noted that these test fail due to rounding errors when the raster is completely symmetrical. In those case cells on the diagonal  can differ, I think in those cases it might just fall in a other grid cell (see also example below). Let me know what you think.

``` r
require(bioRad)
#> Welcome to bioRad version 0.4.0.9261
#> Docker daemon running, Docker functionality enabled
#> No vol2bird Docker image found. Please run update_docker() to download.
require(raster)
#> Loading required package: raster
#> Loading required package: sp
#> 
#> Attaching package: 'raster'
#> The following object is masked from 'package:bioRad':
#> 
#>     bind
data(example_vp)
file.exists( pvolfile <- system.file("extdata", "volume.h5", package = "bioRad"))
#> [1] TRUE
example_pvol <- read_pvolfile(pvolfile)
my_ppi <- integrate_to_ppi(example_pvol, example_vp,xlim=c(-1000,1000),ylim=c(-1000,1000), res=10)
a<-raster(integrate_to_ppi(example_pvol, example_vp, res=raster(my_ppi$data))$data)  
b<-raster(my_ppi$data)
plot(crop(log(a), extent(c(-1000,-900,-1000,-900))))
```

``` r
plot(crop(log(b), extent(c(-1000,-900,-1000,-900))))
```

<sup>Created on 2019-09-15 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


bart1:
Sorry I think i forgot to add the proposed reason for this change. It is to make it easier to do multiple runs onto the same grid avoiding interpolations later.

Issue #198, opened 2019-01-03
time and height alignment of pixels in plot.vpts


adokter:
Pixels are currently
* centralized around the nominal time on the time axis
* centralized around the altitude bin height (HGHT)
However, typical the nominal time is the start time of the acquisition of the polar volume, and HGHT gives the lower altitude of the bin. Each pixels is thus offset by half a pixel in both altitude and time

adokter:
see https://github.com/adokter/bioRad/issues/276 for additional examples by @bart1

bart1:
Thanks for pointing this out I looked for duplicates but could not find them. I'm less sure about the time case, first the acquisition time is not necessarily equal to the interval. Say a complete scan takes 4 minutes but the interval is 15 then the time axis should only be shifted 2 minutes not 7.5. Additionally, typically the first scans are the lower elevation ones (as far as I have seen), these contain most biological information so I'm not sure what is exactly right there.

adokter:
Height offset fixed in #277. Not fixing the time offset for now, as nominal time is ambiguously defined.

PR #277, opened 2019-10-23
Fix height for vp and vpts plot

bart1:
This fixes the height part of #198, it also moves the barbs accordingly

### DIFF
--- a/R/plot.vpts.R
+++ b/R/plot.vpts.R
@@ -202,9 +202,9 @@ plot.vpts <- function(x, xlab = "time", ylab = "height [m]", quantity = "dens",
   # plotdata[is.na2(plotdata)]=(breaks[1]+breaks[2])/2
   # when calculate_vp stdout also differentiates between NA and NaN:
   plotdata[is.na(plotdata)] <- (breaks[2] + breaks[3]) / 2
-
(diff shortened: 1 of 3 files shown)


R/plot.vpts.R
```diff
@@ -220,7 +220,7 @@ plot.vpts <- function(x, xlab = "time", ylab = "height [m]", quantity = "dens",
     if ("ylim" %in% names(args)) {
       h.barbs <- seq(min(args$ylim), max(args$ylim), length.out = barbs_height)
     } else {
-      h.barbs <- seq(x$heights[1], tail(x$heights, 1), length.out = barbs_height)
+      h.barbs <- seq(x$heights[1], tail(x$heights, 1) + interval, length.out = barbs_height)
```
adokter: `interval/2` here as well?

bart1: I think not since this is only giving the start and end of the sequence over which the heights are distributed. Now is it the same as the ylim. Alternatively It could be:
```r
       h.barbs <- seq(x$heights[1] + interval/2, tail(x$heights, 1) + interval/2, length.out = barbs_height) 
```
This would start and end the barbs midway the first and last interval

PR #278, opened 2019-10-23
considerable speedup

bart1:
This updates the functionality to integrate_to_ppi to a raster. Is makes it much faster as it only does one projection of the raster. Note a function scan_to_spdf needed to be added as the converted data is not nessicarly gridded and thus scan_to_raster does not always work


R/scan_convert.R
```diff
@@ -185,3 +185,66 @@ scan_to_raster <- function(scan, nx = 100, ny = 100, xlim, ylim, res = NA, param
   if ("azim" %in% param_to_use) output$azim <- polar_coords$azim
   output
 }
+scan_to_spdf <- function(scan, spdf, param, lat, lon, k = 4 / 3, re = 6378, rp = 6357) {
```
adokter: you added `scan_to_spdf()`, but I think this functionality is already available in `scan_to_spatial()`: https://github.com/adokter/bioRad/blob/e0ede427eb34007dc9985302d40cbdab158e0636/R/scan_convert.R#L13

Please update to using that function

bart1: As I understand it both functions do something considerably different. `scan_to_spatial` takes one scan and converts all measurements to a spatial object (in this case there is no way to influence the definition of the object returned for example when you are only interested in a specific region, or want to annotate a set of points with multiple scans). `scan_to_spdf` takes a spatial points dataframe and maps all attributes of a scan to that spatial points dataframe. I see that the name is rather unfortunate as `scan_to_spatial`  also exists. I used this name since the functionality is very similar to `scan_to_raster` so I tried to stick to the same naming convention.

adokter: > `scan_to_spdf` takes a spatial points dataframe and maps all attributes of a scan to that spatial points dataframe.

I think this is exactly what `scan_to_spatial` does as well, for example see this example:
```
> pvolfile <- system.file("extdata", "volume.h5", package = "bioRad")
> pvol <- read_pvolfile(pvolfile)
> scan_to_spatial(pvol$scans[[1]])
class       : SpatialPointsDataFrame 
features    : 172800 
extent      : -239848.1, 239848.1, -239848.1, 239848.1  (xmin, xmax, ymin, ymax)
crs         : +proj=aeqd +lat_0=56.3675003051758 +lon_0=12.8516998291016 +units=m +ellps=WGS84 
variables   : 9
names       : azim,  range,         distance,             HGHT, DBZH,             VRADH,            RHOHV,               ZDR,             PHIDP 
min values  :    0,    500, 499.980668562476, 213.377999126911,  -30, -23.8809346854687,                0, -15.9058823511004, -178.588235259056 
max values  :  359, 240000, 239848.089357779, 5695.96400438249,   55,  23.6928958296776, 0.99959997474798,  7.90588282048702,  178.588244199753 
```
Maybe you can add a similar example to this thread to illustrate how `scan_to_spdf` is different

adokter: Bit I see your `scan_to_spdf` also takes a SpatialPointsDataFrame as input. Are you overlaying the scan with the points of the input dataframe?

If that's the operation, why not first convert the scan with `scan_to_raster` and then use standard raster package functions to do the overlaying?

bart1: > If that's the operation, why not first convert the scan with scan_to_raster or scan_to_spatial and then use standard raster package functions to do the overlaying?

This would be an option but I would like to avoid doing a secondary interpolation/nearest neighbor matching. This would introduce an extra error. Would your concerns be addressed when `scan_to_raster` would largely rely on `scan_to_spdf`(It needs to be this order since the whole reason to make this function is to avoid the gridded nature of `scan_to_raster` )? I think that should be not too hard and avoid code duplication I did not do this to avoid to involved changes to existing code but with hind sight might have been better.

adokter:
Merged this pull request into a local branch of this repository, https://github.com/adokter/bioRad/tree/bart1-gridSpeedup. @bart1 let's make changes in that branch instead of your external repo, and have the final merge come from this new branch

PR #282, opened 2019-10-25
considerable speedup

adokter:
Continuing https://github.com/adokter/bioRad/pull/278 in a branch native to this repository


R/integrate_to_ppi.R
```diff
@@ -168,21 +168,39 @@ integrate_to_ppi <- function(pvol, vp, nx = 100, ny = 100, xlim, ylim, zlim = c(
...
-    as(scan_to_raster(add_expected_eta_to_scan(x, vp, param = param, lat = lat, lon = lon, antenna = antenna, beam_angle = beam_angle, k = k, re = re, rp = rp), nx = nx, ny = ny, xlim = xlim, ylim = ylim, res = res, param = c("range", "distance", "eta", "eta_expected"), crs = crs, k = k, re = re, rp = rp), "SpatialGridDataFrame")
-  })
+  if(inherits(res,'RasterLayer')){
+    localCrs<- CRS(paste("+proj=aeqd +lat_0=", lat,
+                       " +lon_0=", lon,
+                       " +units=m",
+                       sep = ""
+    ))
+    values(res)<-1
+    spdf <- (spTransform(rasterToPoints(res,spatial=T), localCrs))
```
adokter: This is currently the only use of `scan_to_spdf()`. Here you take your input raster, convert it to a spdf, and transform it to an aeqd projection (`spdf <- (spTransform(rasterToPoints(res,spatial=T), localCrs))`). Wouldn't it be more user-friendly if the new `scan_to_spdf()` takes this raster and CRS as inputs (instead of an spdf that isn't very well defined as input)? Function structure would then be more similar as `scan_to_raster()`.

Also: note that `rasterToPoints()` throws out any raster points with `NA` values, is that the desired behaviour? If so, please explain in documentation what kind of values should be in the input raster of the `res` argument.

R/scan_convert.R
```diff
@@ -185,11 +185,23 @@ scan_to_raster <- function(scan, nx = 100, ny = 100, xlim, ylim, res = NA, param
...
+#' convert a polar scan into a raster
+#'
+#' convert an object of class 'scan' into a raster of class 'RasterBrick'
+#' @inheritParams scan_to_raster
+#' @param spdf an input object of class SpatialPointsDataFrame
+#' @return an object of class SpatialPointsDataFrame. FIXME: not clear what kind of spdf is expected here
+#' @details FIXME: to be added
+#' @export
+#' @examples
+#' FIXME: to be added
```
adokter: Please add roxygen header with title, description, details, and example

R/scan_convert.R
```diff
@@ -185,11 +185,23 @@ scan_to_raster <- function(scan, nx = 100, ny = 100, xlim, ylim, res = NA, param
...
 scan_to_spdf <- function(scan, spdf, param, lat, lon, k = 4 / 3, re = 6378, rp = 6357) {
```
adokter: Let's think of a better name that makes clear you are projecting onto something that you provide as input, not simply a one-to-one conversion of a scan to a spatialpointsdataframe (which is what `scan_to_spatial()` does)

adokter:
moved to issue #293

PR #285, opened 2019-12-26
Fix: Correctly number azimuths in scan plot

barthoekstra:
While manually trying to filter out rays with EM interference, I found out the rays are not numbered correctly when a `scan` is plotted. I'm not entirely sure how `raster()` converts the `x` values, but it should probably scale between 0-1, which it does not.

See the example below.

**Reference PPI**
PPI

EM interference in the PPI is in the 201-202° and 214-215° rays, which matches nicely with the raw data. However, plotting the scan with the incorrect code shows the interference shifted much closer to 155°.

| Incorrect `scan` plot | Corrected `scan` plot |
| --- | --- |
| Incorrect_Scan |  Corrected_Scan |

### DIFF
--- a/R/plot.scan.R
+++ b/R/plot.scan.R
@@ -70,7 +70,7 @@ plot.scan <- function(x, param, xlim = c(0, 100),
   # convert to points
   dimraster <- dim(data)
   data <- data.frame(rasterToPoints(raster(data)))
-  data$x <- (1 - data$x) * dimraster[2] * c(x$attributes$where$nrays) / 360
+  data$x <- (data$x / max(data$x)) * dimraster[2] * 360 / c(x$attributes$where$nrays)
   data$y <- (1 - data$y) * dimraster[1] * c(x$attributes$where$rscale) / 1000
   # change the name from "layer" to the parameter names
   names(data) <- c("azimuth", "range", param)


adokter:
Hi @barthoekstra, I will be pushing a revised version of plot.scan very soon from develop to master, which was already quite different from the one you made edits to. Please verify of this issue persists

adokter:
Please try the latest bioRad dev version (0.5.0.9289) and let me know if the issue persists:

install instructions here:
http://adriaandokter.com/bioRad/dev/

barthoekstra:
Thanks for the update. The problem still persists, but can be solved by e.g. flipping the raster along the `y` direction.

Change line:
```r
data <- raster::as.data.frame(raster(t(data), ymn=0,ymx=360, xmn=0, xmx=rscale*dimraster[1]), xy=T)
```

To:
```r
data <- raster::as.data.frame(flip(raster(t(data), ymn=0,ymx=360, xmn=0, xmx=rscale*dimraster[1]), direction = "y"), xy=T)
```

adokter:
Thanks @bart1, fixed by commit https://github.com/adokter/bioRad/commit/721835618a4967a2ff2612c50e9e82daccfcefd5

PR #337, opened 2020-04-08
Tests for rcs and rcs<-

peterdesmet:
This PR is a test for the tests hackathon we plan to do next week.

## Functions

- `rcs()` for vp, vpts, and list of vp
- `rcs()<-` for vp, vpts, and list of vp

## Tests

- Get error for incorrect `x` in `rcs(x)`
- Get error for incorrect value in `rcs(x) <- value`
- Output should be double (for 3 classes)
- Output should be 11 by default (for 3 classes)
- Output should be same as assigned new value (for 3 classes)

## Not covered

- [x] Class of object remains the same when assigned a rcs, e.g. `rcs(vp) <- 2` => `vp` is still a `vp` object. This seemed excessive to test.
- [x] `vpi` classes
- [x] `Input must be list of vp objects.`
- [x] `x$attributes$how$rcs_bird <- value`
- [x]  This method also updates the migration densities in `x$data$dens`:`x$data$dens <- x$data$eta / value`
- [x] Threshold for sd_vvp is not set, defaulting to 2 m/s: `x$attributes$how$sd_vvp_thresh <- 2
    x$data$dens[x$data$sd_vvp < 2] <- 0`


peterdesmet:
I just notice this function is also listed in the namespace as working for `vpi`. I will update the definition of the value and add more tests.

peterdesmet:
Added more tests now to cover true functionality of the function + vid. See:

https://github.com/adokter/bioRad/blob/6cd096731b95b4901328ea7473b3b08bb2155499/tests/testthat/test-rcs.R

@adokter ready for review: especially have I organized the tests

adokter:
Do we want explicit tests for nonsense input arguments, like I added in c7cc0f0. Problem with assert_that errors is that we cannot be sure that those will change, so hard to test on the contents of the error message. But simply testing for an error like now is also rather simplistic.

peterdesmet:
@damianooldoni do you have any answer to Adriaan’s question above.

@adokter I think we should revert https://github.com/adokter/bioRad/commit/2fe5b5d9aa0bca5a07da174c56f7bb679a88d144: i want to test it contains any value >0 so we van test it is set to 0 later

damianooldoni:
Hi @adokter, @peterdesmet . I like to have always `assert_that` errors in my functions with a specific message. Defensive programming is something I like a lot: as user of a package I like when the error is in the very beginning and it is documented then a rare error message in the middle of a complicated function. So, I like to spend a little time to write an short but clear error message.
Typically two assertion errors:
1. about type (must var be a char? numeric? data.frame?)
2. about value (must be greater than 0? string longer than 1? nrow(df) > 0? Does column `a` exist in `df`?

peterdesmet:
And here's a good tutorial on how to phrase error messages: https://style.tidyverse.org/error-messages.html

PR #342, opened 2020-04-09
Enable Roxygen markdown

peterdesmet:
Part of #240 

Note, this just enables it + fixes some conversion errors (like avoiding square brackets since these turn into links). Most of the documentation is still regular Roxygen.


peterdesmet:
The option to ignore whitespace changes came in really handy here 😄:

https://github.com/adokter/bioRad/pull/342/files?w=1

PR #347, opened 2020-04-14
Setup test files for all functions

peterdesmet:
This PR sets up (empty) test files for all functions. Also renamed to 2 `.R` files to their function name (`filter` and `sd_vvp`). Hoping to get this into develop branch by tomorrow, so we can start. 😉 

- @bart1 I assume `test-raster_input.R` is for the `scan_convert` family of functions? Ok if I move it over to `test-scan_convert.R`?
- @bart1 @adokter what are the files `is.nan.data.frame.R` and `hooks.R` files for? Are those helper functions? Can we move those to a `zzz.R` file (seems to be a convention).


adokter:
* `is.nan.data.frame` returns a logical matrix for a data.frame at each location where a `NaN` value occurs. It's only used under the hood, so indeed a helper function
* `hooks.R` contains several hooks for namespace events, i.e. functions executed when the package is loaded or attached. These are used to check for the presence of an active Docker daemon, setup some global variables, and to print a startup message.

adokter:
`test-raster_input.R` contains tests for `project_as_ppi()`, `integrate_to_ppi()` as well as some `scan_convert()` operations, so I think we need to break it up over several of these.

adokter:
I've broken up `test-raster_input.R` in https://github.com/adokter/bioRad/pull/347/commits/058c220c93213a975505c133b06d80cba72d7041

adokter:
Merging now, so I can start writing a few tests

PR #348, opened 2020-04-15
unit tests test-pvolfile.R

adokter:
unit tests for:
* `is.odimfile()`
* `get_odim_object_type()`
* `is.pvolfile()`


adokter:
punctuation and documentation needs to be checked still.

adokter:
Code coverage is only 61.9% because the function code includes several checks for incorrectly formatted hdf5 files (that are strictly not fully ODIM compliant, but might be encountered nonetheless). Using only example files such exceptions can't be tested.

peterdesmet:
@adokter see my 2 commits for shorter tests + updated documentation style

peterdesmet:
@adokter this PR fails because it tests `is.odimfile()` while that function is not in the namespace. Should it be?

adokter:
`is.odimfile` doesn't have to be in namespace, it's functionality is covered by higher-level `is.pvolfile`. I corrected the namespace by changing to `bioRad:::is.odimfile`

adokter:
\document

PR #351, opened 2020-04-15
Get_scan unit tests and error message

CeciliaNilsson709:
Changed from stopifnot to assert_that to be able to have a better error message, what do you think?


peterdesmet:
My commits don't appear here yet because of GitHub status failure. Will merge once GitHub is again fully operational.

peterdesmet:
Also added an `assert_that` for `elev` = number

PR #357, opened 2020-04-17
Reformatted the whole package with styler

niconoe:


peterdesmet:
Nico, I won't accept this PR as I would have to resolve too many conflicts with pending PR. We should run it on the develop branch directly at a later point.

PR #365, opened 2020-04-20
adding volume coverage function

bart1:
Currently there is no function to calculate volume coverage patterns. I made a first draft including some tests. It tabulates all possible attributes and outputs a selection by default. It does not change the name of the attributes which might be tricky with multiple naming conventions but it will still give a quick overview

One test fails as the order of parameters seems to differ between `example_pvol` and `example_scan` ? is that on purpose or because the read function now does something different from when `example_scan.rda` was created?

Some considerations, currently the result does not include any geographic information. it seems not necessary for this function as this is more something for the whole pvol level and here the focus is on volume coverage.


bart1:
@adokter , I think the function does what I want it to do I just fixed a typo in the examples. 
One test fails due to this difference:
```r
> vcp(example_pvol)[1, ]
   how.NI how.highprf how.lowprf where.elangle where.nbins where.nrays where.rscale                         params
 24.06897         600        450           0.5         480         360          500 DBZH, VRADH, RHOHV, ZDR, PHIDP
> vcp(example_scan)
   how.NI how.highprf how.lowprf where.elangle where.nbins where.nrays where.rscale                         params
 24.06897         600        450           0.5         480         360          500 VRADH, DBZH, ZDR, RHOHV, PHIDP
```
It looks like the order of parameters is different for the example pvol and example scan, do you want to adjust the test or change the data?

adokter:
@bart1 the test will have to be adjusted, because the metadata in ODIM files is highly variable, much of it is even optional, therefore you see huge variation in what countries store under `/what`, `/where` and `/how` attribute groups. 

For example, PRF is sometimes missing.

Note also that France has a triple-PRT scheme, so they have three pulses coming at different intervals (instead of burst of pulses with one frequency), so they might list three PRTs (pulse repetiton times) instead of the standard two PRFs.

Also, vcp = volume coverage pattern, so it is a property of the polar volume (creating a VCP for a single scan (as listed in the documentation) isn't really meaningful). I suggest having the vcp method work only on a pvol object.

bart1:
@adokter most of the function works pretty generic and does not really care about the naming of attributes, it tabulates all (see `select=NULL`). it just returns a common subset to keep things more manageable in most cases, for french data mid prf is returned when present. It working on a scan is mostly a side effect of implementing it recursively. I think it might also be useful for identifying differences between scans. The function could be renamed to something like `coverage_table` to avoid volume in the naming.

bart1:
@adokter any thoughts on my last comment here should we rename it to `coverage_table` or `coverage_pattern`? I just recently noticed I was again missing the function.

adokter:
Hi, @bart1 

- how about naming it `attribute_table()`? Because essentially you take the `x$attributes` and make it into a table. In that way we keep it more generic and you can generalize it also to `vp` and `vpts` objects.
- Can we change the `.` separating the attribute group and the attribute into a `/`, in that way we keep it more aligned with the ODIM hdf5 convention.
- in `nyquist_velocity()` we have `prf1` and `prf2`, while here we have `lowprf` and `highprf`. Renaming the parameter in nyquist_velocity is probably the most consistent, as again we stick to ODIM hdf5 (there is also a `midprf` relevant for French data).

bart1:
I just renamed to `attribute_table` that seems good to me.

I'm slightly hesitant to change to `/` as it is a very strange character for column names in R. Columns forexample cant be addressed like normal:
``` r
d<-data.frame(b.b=1)
d$'a/a'<-2
d$b.b
#> [1] 1
d$`a/a`
#> [1] 2
d$a/a
#> Error in eval(expr, envir, enclos): object 'a' not found
```

Therefore I would opt for keeping the `.`.

Shall I directly rename `nyquist_velocity()`?

adokter:
Ok got it with the confusing divide operator. Throughout we've been using `_` to separate words following http://adv-r.had.co.nz/Style.html, any issue with using that instead of `.`?

adokter:
* rownames attribute needs a fix, should match the scan index:
```
rownames(attribute_table(pvol))
[1] ""  "1" "2" "3" "4"
```
* rename `params` -> `param` (we deprecated the use of plural params earlier in other functions)
* do we really need the `add_params` argument? it's easy for users to discard a column anyways.
* generalize to include `vp` and `vpts`
* can we return all attributes when `select='all'`

Finally, I wonder whether it wouldn't be better if we report a transposed version of the table, that would make more sense with `vpts` and `vp` objects that don't have multiple records, like a `pvol` with multiple `scan`. And to have the attribute group and attribute name in separate table columns, like this:

|   | group | attribute | value1 | value2 | value3 | scan... |
|---|-------|-----------|-------|-------|-------|---------|
|   | how   | NI        | 8     | 8     | 8     |         |
|   | where | elangle   | 1     | 2     | 3     |         |
|   | where | nrays     | 360   | 360   | 720   |         |
|   | where | nbins     | 1000  | 1000  | 2000  |         |

So for a `pvol` the `valuei` refer to scans. For `scan`, `vpts`, `vp` we have only one value column.

Then we don't have the naming issue with `.` and `_`, keep the attribute structure, and you get a sensible table structure also for vp and vpts.

bart1:
I have not updated the function by using the column name `param` although it generally contains multiple. I am not so fond of the transpose as binding multiple tables then would produce different scans to be either pasted below each other although they are hugely different or you get a very wide table. I think it is more clear if each row represents a scan.

bart1:
@peterdesmet while fixing this PR I noticed the order of parameters between `example_scan` and `example_pvol` differ (for the same scan) is this something you would like to have consistent and change? or shall I not test for it?

```r
> names(example_scan$params)
[1] "VRADH" "DBZH"  "ZDR"   "RHOHV" "PHIDP"
> names(example_pvol$scans[[1]]$params)
[1] "DBZH"  "VRADH" "RHOHV" "ZDR"   "PHIDP"
```

peterdesmet:
@bart1 regarding https://github.com/adokter/bioRad/pull/365#issuecomment-857739062, I guess it would be better if they are consistent, but where would we enforce it? @adokter does the order matter?

bart1:
> @bart1 regarding [#365 (comment)](https://github.com/adokter/bioRad/pull/365#issuecomment-857739062), I guess it would be better if they are consistent, but where would we enforce it? @adokter does the order matter?

I would not enforce it but thought it might be nice to have it at least consistent in the example data. I could also adjust the tests accordingly

adokter:
@bart1 @peterdesmet what do you think of reformatting the table as suggested in https://github.com/adokter/bioRad/pull/365#issuecomment-667251163. The `where.attribute` and `how.attribute` names with the dots are a bit of a style break, and it's nicer to have the where and how as their own table column

adokter (approved):

PR #369, opened 2020-04-27
Move code for as.data.frame to own file

peterdesmet:
Move test as well


peterdesmet:
Does not include changes yet to make as.data.frame() a generic function with just single documentation, since I'm a bit stuck there. See #369

Issue #355, opened 2020-04-16
dimension of bioRad classes reported by dim()


adokter:
Subsetting the example data currently works as follows:

```
# vp
example_vp$data[height,quantity]

# vpts option 1
example_vpts$data[[quantity]][height, time]
# vpts option 2
example_vpts[time]$data[height, quantity]

# vpi
example_vpi <- integrate_profile(example_vpts)
example_vpi[time,quantity]

# param
example_param[range,azim]

# scan
example_scan$params[[param]][range,azim]

# pvol
example_pvol$scans[[scan]]$params[[param]][range,azim]
```
For the example data we have the following index maxima:
```
height = 25
time = 1934
quantity = 15 of 16
scan = 3
param = 5
range = 480
azim = 360
```
Checking `dim()`:
```
dim(example_vp)    #> c(25,16)      correct
dim(example_vpts)  #> c(25,1934,15) incorrect, given above subsetting you expect c(15,25,1934) or c(1934,25,15)
dim(example_vpi)   #> c(1934, 12)   correct
dim(example_param) #> c(480,360)    correct
dim(example_scan)  #> c(5, 480,360) correct
dim(example_pvol)  #> NULL          incorrect, should be 3
```
Suggesting the following changes:
* `dim(example_vpts)` should give `c(1934,25,15)` i.e. `c(time, height, quantity)`, which follows the vpts subsetting order of option 2, and is the most similar in order to `dim(example_vp)`
* `dim(example_pvol)` should give `c(3)` i.e. the number of scans only. We can't give the deeper dimensions, as they may be different for various scans.

peterdesmet:
Fixed in commits referenced above

PR #372, opened 2020-04-27
Add tests for summary, print, dim, subset, is.object, c.vp()

peterdesmet:
Adds tests for object functions for:

- pvol
- scan
- param
- ppi
- vp
- vpts

Fixes:

#370 
#355 

Cannot get full code coverage for vpts, mainly due to this code, which I don't full understand:

https://github.com/adokter/bioRad/blob/4eac9cedeee915b9577bfb183ff63bf02d733414/R/vpts.R#L162-L178


adokter:
> Cannot get full code coverage for vpts, mainly due to this code, which I don't full understand:

That piece of code deals with vpts of length two, and negative subset indices, which remove profiles. The code makes sure that the correct vp object is returned, ie for -1 the profile at index 2, and for -2 the profile at index 1

This relates to the different discussion if it's a good idea to have changes in the class output depending on input parameters (typically not, but here it happens), and whether we shouldn't deprecate class vp, or make class vp heritable from class vpts

peterdesmet:
Right, in that case you might want to revert https://github.com/adokter/bioRad/pull/372/commits/143c3f32fb3f4d3ee3962d2cf3dbba914665eb7f or is this unrelated?

adokter:
I think it's ok to leave `143c3f3` in place and allow creation of empty vpts objects, such that we we both allow example_vpts[0] and example_vpts[c()]

PR #393, opened 2020-05-21
Composite PPIs for multiple parameters at once

barthoekstra:
Still a work in progress, especially with unit tests and documentation missing, but let me know what you think of the following changes I've made so far. Feedback on code is also highly appreciated, as I only started using R this year and may be doing some unconventional things here and there… ;-) 

**Multiple `param`'s and `method`'s**
Parameters can be selected like in `read_pvolfile()` by giving a list of names, or simply supplying `param = "all"`. Similarly, a single `method` can be set to determine the processing for _all_ PPI parameters, but they can also be set separately for each of the parameters by providing a list with method names of the same length as the list of parameters.

**`res` in meters by default**
After using `integrate_to_ppi()` or even `project_as_ppi()` I assumed the `res` parameter in the composites would work similar and set a resolution in meters, but instead they were (by default) longlat raster units. These are quite unintuitive to work with, I think. I've instead implemented that if no `crs` is provided but a `res` is, a common raster grid with meter units with the origin at the mean latlon location is used for the composites.

I'm als thinking of adding some optional 'overlap' PPI param to the output, to visualise/filter where composite param values are a product of multiple PPIs.


barthoekstra:
I have added a `coverage` parameter to get an indication of the number of overlapping `ppi` pixels a composite pixel ‘contains’, or a list of the corresponding ODIM radar IDs. This is something I personally need, but I admit that it may be quite a 'niche' feature? Although, as the composite PPI is always on a rectangular grid, it can help to distinguish between areas where there could be datapoints vs where not...

Additionally, as the `ppi` objects do not contain any information about the maximum range of the radar, currently the area of coverage is always rectangular. This is fine if the range of the each of the composited PPIs is smaller than the maximum range of the respective radars, but otherwise the rectangular shape is not very useful.

adokter (commented):
Hi @barthoekstra, looking good. The only issue I have is with the coverage="radars" option, because this breaks plotting of the coverage quantity. Since the function outputs a ppi, I'd like to keep it consistent with other ppi objects

So my suggestion would be to only include coverage="count", and since there is only one type of coverage output it can become a TRUE/FALSE flag instead.

adokter:
I've also changed to param='all' as the default

barthoekstra:
Ok. Fixed that.

R/composite_ppi.R
```diff
@@ -80,19 +88,25 @@ composite_ppi <- function(x, param = "DBZH", nx = 100, ny = 100, xlim, ylim, res
...
-  }
-  else {
-    crs <- CRS("+proj=longlat +datum=WGS84")
+    t_crs <- CRS(as.character(raster::projection(crs)))
+  } else {
+    t_crs <- NULL
   }
+  if (!all(method %in% c("max", "min", "mean", "idw"))) stop("'method' should be one or multiple of 'max', 'mean', 'min' or 'idw'")
+  if (length(method) != length(param) & length(method) != 1) stop("'method' should be of length 1 or length(param)")
+  if (!is.logical(coverage)) stop("'coverage' should be a logical")
```
bart1, line 103: @barthoekstra : A quick hint this can easier be written as `assert_that(is.flag(coverage))`. It directly ensures the length is 1

adokter (approved):

R/composite_ppi.R
adokter: fixed this in b89d6c37c7f96c4f05ee4d8588328a3f70b0612b

PR #435, opened 2021-05-28
Astart rstart

### LINKED ISSUE #434: implementing astart and rstart
Ok I looked in this a bit further, also reading the odim specs (https://www.eumetnet.eu/wp-content/uploads/2019/01/ODIM_H5_v23.pdf). The fix you propose and I had in my head is the most obvious one and would work place the point in the center of the range gate for most cases. 

There are however two related issues. Some radars seem to use the `rstart` attribute. I did a quick scan over the baltrad radars (one per country, first scan)
```
        a1gate   elangle nbins nrays   rscale rstart
BE/JAB/    137 0.5000000   300   360  500.000  0.000
CH/ALB/      0 0.3790341   420   360  499.998  0.000
CZ/BRD/    190 0.1000000   650   360  400.000  0.000
DE/ASB/    100 0.8020020   180   360 1000.000  0.000
DK/BOR/    313 0.4779050   472   360  500.000  0.500
EE/HAR/    330 0.5000000   830   360  300.000  0.900
ES/ALM/    394 0.4998779   240   450 1000.000  0.125
FI/ANJ/    188 0.3000000   500   360  500.000  0.000
FR/ABB/    227 0.4000000   267   360  960.000  0.000
HR/BIL/    184 0.5000000   241   360 1000.000  1.000
NL/DHL/     83 0.3000000   838   360  223.500  0.000
PL/BRZ/      0 0.5000000   250   360 1000.000  0.000
PT/FAR/    194 0.1000000   200   720  500.000  0.000
SE/ANG/    354 0.5000000   480   360  500.000  0.000
SI/LIS/    293 0.5000000   249   360 1000.000  0.000
SK/JAV/    192 0.0000000   960   360  250.000  0.000
```
It looks like some countries use this attributes but biorad does not account for it. I visually checked PPI for `DKBOR` and `EEHAR` with some wind turbine locations and applying the proposed shift would probably make the data more plausible.

Similar there is an attribute `astart` the french radar `FRABB` seems to use this (-.5 degrees). Which if i read the docs correctly mean the whole image needs to be rotated half a degree. Looking at some plots for this it might be indeed the better fit.

The fixes for these later two problems are however probably more elaborate as they would have implications throughout the package. (each time going from Cartesian to polar and vise versa).

A few radars also report start and end azimuths for each ray. The deviations from the implied values are however not very big (about 0.02 degrees). I think trying to this correctly is probably a bit over the top (given beam width usw).

_Originally posted by @bart1 in https://github.com/adokter/bioRad/issues/430#issuecomment-849752250_

bart1:
this fixes #434 I opted not to set `rstart` and `astart` to 0 in the `geo` object as I think when it is not set in the odim file it should be null and not 0. I directly multiply it when reading to keep in the package using the range  in meters and not mix it up with KM.

I think I applied the fixes in all places I found it relevant (searching where `ascale` , `rscale` and `polar_to_index` is used)

@adokter did I miss some place?

### DIFF
--- a/R/project_as_ppi.R
+++ b/R/project_as_ppi.R
@@ -190,14 +190,19 @@ sample_polar <- function(param, grid_size, range_max, project, ylim, xlim, k = 4
   # get scan parameter indices, and extract data
   index <- polar_to_index(
     cartesian_to_polar(coordinates(gridTopo), elev, k = k, lat = attributes(param)$geo$lat, re = re, rp = rp),
-    attributes(param)$geo$rscale,
-    attributes(param)$geo$ascale
+    rangebin = attributes(param)$geo$rscale,
+    azimbin = attributes(param)$geo$ascale,
+    azimstart = ifelse(is.null(attributes(param)$geo$astart), 0, attributes(param)$geo$astart),
+    rangestart = ifelse(is.null(attributes(param)$geo$rstart), 0, attributes(param)$geo$rstart)
   )
   # set indices outside the scan's matrix to NA
   nrang <- dim(param)[1]
   nazim <- dim(param)[2]
   index$row[index$row > nrang] <- NA
   index$col[index$col > nazim] <- NA
+  # rstart can result in locations outside of the radar scope close to the radar
+  index$row[index$row < 1] <- NA
+  stopifnot(all(index$col >= 1))
   # convert 2D index to 1D index
   index <- (index$col - 1) * nrang + index$row
   data <- as.data.frame(param[index])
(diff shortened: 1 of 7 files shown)


R/project_as_ppi.R
```diff
@@ -190,14 +190,19 @@ sample_polar <- function(param, grid_size, range_max, project, ylim, xlim, k = 4
...
+    rangestart = ifelse(is.null(attributes(param)$geo$rstart), 0, attributes(param)$geo$rstart)
   )
   # set indices outside the scan's matrix to NA
   nrang <- dim(param)[1]
   nazim <- dim(param)[2]
   index$row[index$row > nrang] <- NA
   index$col[index$col > nazim] <- NA
+  # rstart can result in locations outside of the radar scope close to the radar
+  index$row[index$row < 1] <- NA
+  stopifnot(all(index$col >= 1))
```
adokter, line 205: could you add an informative error message that explains why this stop happens and what to do. Is this when rstart has a value that makes certain ranges negative?

adokter (changes requested):
Thanks @bart1, looking good. Could you add a more informative error message at the stopifnot statement?

R/project_as_ppi.R
bart1: I think this error basically should not occur it is basically a sanity check. The only reason I could occur I could think of is when some integer float conversions result in just negative numbers (rounding errors). I can also remove it. Or just write that. I think there is no consistent reason why this should occur

PR #439, opened 2021-06-07
Test dbz eta

Rafnuss:


R/dbz_eta.R
```diff
@@ -20,6 +20,9 @@
 #' dbz_to_eta(7, 5) / 11 # C-band
 #' dbz_to_eta(7, 10) / 11 # S-band
 dbz_to_eta <- function(dbz, wavelength, K = 0.93) {
+  assert_that(class(dbz) == "numeric" & all(dbz>0), msg = "`dbz` must be a strictly positive numeric.")
```
peterdesmet, line 23: Perfect to use `assert_that`. In error message I would use 'single quotes' to indicate literal values, so 'dbz', but we likely haven't been consistent.

tests/testthat/test-dbz_eta.R
```diff
@@ -1,8 +1,24 @@
 test_that("dbz_to_eta() returns error on incorrect parameters", {
+  test_that("dbz_to_eta() returns error on incorrect parameters", {
+    expect_error(dbz_to_eta('not_a_dbz', 10))
```
peterdesmet, line 3: You can also indicate what error message the user will get back: 

```R
expect_error(dbz_to_eta('not_a_dbz', 10), "'dbz' must be a strictly positive numeric.", fixed = TRUE)
```

If you drop the `fixed = TRUE` you can write part of the error message that needs to match:

```R
expect_error(dbz_to_eta('not_a_dbz', 10), "'dbz' must be")
```

peterdesmet (changes requested):
Minor textual changes

R/dbz_eta.R
peterdesmet: Actually, ignore that. The styleguide says backticks `...`: https://style.tidyverse.org/error-messages.html so lets use that.

tests/testthat/test-dbz_eta.R
Rafnuss: Ok, nice!  should be resolved with 3f45a0f743dd35f99641b8bc71cc8c576317f034

adokter:
Hi @Rafnuss, the tests were failing because you didn't allow dbz to be negative (it's a log quantity)

I also pushed some changes to allow K and wavelength to be vectors, no reason not to allow this.

adokter (approved):

PR #445, opened 2021-06-08
convert_legacy() doc + tests

peterdesmet:
- Update doc
- Shorten example
- Add tests:
  - Input parameter
  - Output = same object type
  - Output is a modern bioRad object


CeciliaNilsson709 (commented):
looks good

PR #448, opened 2021-06-09
test-download_vpfiles

Rafnuss:


Rafnuss:
Should we test that the radar code exist ? If so, what's your recommended option:
- read the file with radar code https://lw-enram.s3-eu-west-1.amazonaws.com/radars.csv? Probably slow down the function for not much?
- Read the file only if download return of 404 error?
- Store all radar code somewhere in the function to test?

Rafnuss:
> Should we test that the radar code exist ? If so, what's your recommended option:
> 
> * read the file with radar code https://lw-enram.s3-eu-west-1.amazonaws.com/radars.csv? Probably slow down the function for not much?
> * Read the file only if download return of 404 error?
> * Store all radar code somewhere in the function to test?

It's actually pretty quick to read (compared to downloading the vpts files), so I guess it should be ok as implemented in  c067bd7e4d01db64535839ef26686add6061b0ea. let me know what you think.

Rafnuss:
@peterdesmet, Can you help me with the error left. I don't understand the problem with https://github.com/adokter/bioRad/blob/master/vignettes/range_correction.Rmd. I can't find any `assertthat `on this code and I don't see that `download_vpts `has been used there. You mentioned something about merging it to the master branch to solve the issue?

adokter:
> @peterdesmet, Can you help me with the error left. I don't understand the problem with https://github.com/adokter/bioRad/blob/master/vignettes/range_correction.Rmd. I can't find any `assertthat `on this code and I don't see that `download_vpts `has been used there. You mentioned something about merging it to the master branch to solve the issue?

@Rafnuss the remaining error popped up a few days ago and isn't related to your code, I believe it's an issue with a dependency, not something we need to worry about right now

adokter:
I noticed the `radar` keyword was changed to `radars`, my suggestion would be to revert that, as it breaks backward compatibility, and quite often the use case is to download data for a single radar.

Rafnuss:
> I noticed the `radar` keyword was changed to `radars`, my suggestion would be to revert that, as it breaks backward compatibility, and quite often the use case is to download data for a single radar.

I only made this change in the doc example, the code in itself was already using `radars` (see https://github.com/adokter/bioRad/blob/28f5d978574c05bfe374d05c411d04ebaad6664b/R/download_vpfiles.R#L39). Can this example work currently with this difference in R?

adokter (approved):

Issue #414, opened 2020-11-10
get_scan and multiple similar elevation scans have same angle


bart1:
Currently `get_scan` returns "Returns the scan (scan) from a polar volume (pvol) with elevation angle closest to elev.". Some radars have multiple scans at the same elevation. It is kind of ambiguous what is currently returned in those cases. I think the might be a few options, I guess the current implementation returns the first one with the minimal difference. In this case it might be worth to issue a warning about other scans being available with the same height. Alternatively multiple scans might be returned as a list that match the criteria. This might be a breaking  change for some people. @adokter what do you think?

adokter:
Hi @bart1, I agree to avoid breaking changes the default should remain returning only a single scan is returned. I like the idea of giving a warning when multiple scans have the same elevation angle. How about adding a third (optional) boolean argument `all` (default: `FALSE`) that when `TRUE` returns a list of scans if multiple scans are present at the same elevation?

PR #450, opened 2021-06-09
fix 414

bart1:
This fixes the issue #414 the only possible disadvantage is the behaviour is the same if there are two scans with the same elevation angle  (e.g. 1 and 1 degree) or if you try to retrieve an elevation angle exactly in between two other ones e.g. `elev=2` while elevation angles 1 and 3 exist. I to always return a list as it is probably easier to work with compared to when the type returned varies.


adokter:
@bart1 what is the most common use case for `all=TRUE`?

1. To retrieve multiple scans with the same elevation (like 1 and 1)
2. or to retrieve all scans equally close (like 1 and 3 when requesting 2).

If 1. is the most common use case, I think it might be easier for the user to simply return the lowest elevation angle scan when the requested elevation is right in between two scans.

adokter (changes requested):
Depending on most common use case, suggest to return the lowest elevation scan when requested elevation scan is right in between two scans. This might occur quite often, and it might be more informative for the user if the length of the list indicates how many scans there are at the same requested elevation angle

bart1:
> @bart1 what is the most common use case for `all=TRUE`?
> 
>     1. To retrieve multiple scans with the same elevation (like 1 and 1)
> 
>     2. or to retrieve all scans equally close (like 1 and 3 when requesting 2).
> 
> 
> If 1. is the most common use case, I think it might be easier for the user to simply return the lowest elevation angle scan when the requested elevation is right in between two scans.

I'm not a 100% sure. My main problem was with case 1 as the dutch radars have 3 scans at the same elevation. I do not use if for other applications. So I'm happy to return the lowest in case of conflict and not warn in that case?

adokter:
>  I'm happy to return the lowest in case of conflict and not warn in that case?
Ok let's do that, I'll update the PR with that change

adokter (approved):

Issue #343, opened 2020-04-09
binding profiles with different altitude layer definitions into a vpts


adokter:
Will require functions to align the altitude layers of profiles, and perform interpolations where necessary. Requires expanding the functionality of `bind_into_vpts()`

adokter:
See also issue #341 reported by @pecard for a reproducible example of where bind_into_vpts() function currently fails.

adokter:
see also https://github.com/adokter/bioRad/issues/341#issuecomment-617401758 comment by @bart1

adokter:
This issue is partially fixed by PR #457, which adds support for profiles with different maximum altitudes. What remains is
* profiles with different altitude bin widths
* profiles with misaligned altitude bin heights

adokter:
see suggestion by @bart1 in https://github.com/adokter/bioRad/pull/698#issuecomment-2742706353

PR #457, opened 2021-06-10
bind profiles with different number of altitude layers (#343)

bart1:
This should make it possible to combined profiles with different heights as long as they are aligned. I have not adjusted the `maxheight` attributes as I interpreted this as the height for which it has been calculated and not interpolated

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 * Warn when more then one scan could be returned (#414) and provide alternative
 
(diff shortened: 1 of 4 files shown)


adokter (approved):

bart1:
> Hi @bart1, I added updating of the `attributes$where$levels`, since that is an attribute that is required (maxheight isn't always present, so leaving that for now)

Thanks sorry missed that one

PR #461, opened 2021-06-16
Test plot functions

CeciliaNilsson709:
Unit tests for plot functions, mostly using the vdiffr package and "snapshot" function in testthat.  First time the tests run they creates a snapshot, stored under testthat/_snap/function_name. Then each time test is run it creates new snapshot and checks that it looks identical to previously saved snapshot. If it fails, a warning is given and the new snapshot is stored with ".new" appended to name for manual inspection.


adokter:
FYI I got the following warning to update R:
```
Please update to R 4.1.0 and regenerate the vdiffr snapshots.
```

peterdesmet (approved):
Looks excellent, improves coverage by 14% 🥳 . Made some cosmetic changes.

Oddly enough, clicking `Run tests` on the file seems to work, while `devtools::test()` removes all snapshots. Seems related to https://github.com/r-lib/testthat/issues/1393. Even with new dev version of `vdiffr` the issue remains.

Should we wait before merging?

peterdesmet:
The removal of the `_snaps` only happens on `devtools::test()`, not on `devtools::check()` which is what the CI is running. So I think it is safe to merge this. The worst thing that can happen is that someone runs `devtools::test()` locally and sees in GitHub Desktop that all the snaps have been removed. These should then not be committed. If they are, we will notice in a PR + we can always recreate and commit them. Hopefully, `devtools::test()` will get updated at some point so this doesn't occur anymore.

PR #444, opened 2021-06-07
user-defineable color scale in plot.vpts()

adokter:
Adds arguments to plot.vpts()
* `palette`: a user-defined colorscale
* `na_color`: the color to plot NA values (gray by default)
* `nan_color`: the color to plot NaN values (default lowest value of color scale)

Necessary step for fixing #424

PR #470, opened 2021-06-29
Add write_pvolfile

barthoekstra:
Cleaned the code up and added tests. Tested it on some Dutch and German pvols and it worked fine, but if there are countries with somehow ‘different’ polar volume data that's still worth checking, let me know.


adokter:
thanks @barthoekstra, somehow the PR got mixed up with #444 so continuing this PR in #471, closing here

PR #471, opened 2021-06-29
add write_pvolfile (see #470)

adokter:
replaces PR #470


adokter:
@barthoekstra looking good! I noticed NEXRAD files (available here https://s3.amazonaws.com/noaa-nexrad-level2/index.html) are stored in a considerably larger size than the original. This is because the `nexrad_to_odim()` function doesn't track the original encoding and stores everything as 64-bit floating point, something that could be improved.

Is there an easy way to force `write_pvolfile()` to write up to a certain bit size. E.g. a 16 bit floating point would do the trick for most radar quantities and would be a good default.

barthoekstra:
Is there some way to reliably identify NEXRAD specific polar volumes? I presume through the fields in the source attribute, but you'll know better than me what a robust identifier would be.

You could then write NEXRAD files using
```
h5createDataset(fid, dataname, dim(pvol$scans[[i]]$params[[j]]), H5type = "H5T_IEEE_F32LE")
```

You can consider replacing `H5T_IEEE_F32LE` with `H5T_STD_I16LE` or any of the other types listed [here](https://support.hdfgroup.org/HDF5/doc/RM/PredefDTypes.html) (or by running `h5const("H5T")`) to shave off even more of that filesize. If I understand it right — it's no straightforward reading material — there are no standard 16-bit floating point datatypes in the HDF5 spec.

adokter:
Now storing 64-bit floats as 32-bit floats for NEXRAD files, but this results only in a modest reduction of file size. Large file size for NEXRAD needs to be addressed in adokter/vol2bird#180

PR #472, opened 2021-06-30
Update integrate_profile.R

Rafnuss:
The main aim of this request is to account more accuratly for the bins overlaping with `alt_min` or `alt_max`.
This is done by creating an interval value (\Delta h) for each bin (instead of a constant troncated below alt_min and above alt_max). 
Its value is zeros for bins completly below `alt_min` or above `alt_max`, the constant `interval` value (e.g. 100 or 200m) for all bin completly above alt_min and below `alt_max`, and **a ratio of interval** for the bins overlaping `alt_min` and `alt_max`.


R/integrate_profile.R
```diff
@@ -136,6 +115,28 @@
...
+#' the height-averaged v component (south to north) is defined as:
+#'
+#' \deqn{v = \sum_i dens_i v_i / \sum_i dens_i}{v = \sum_i dens_i v_i / \sum_i dens_i}
+#' }
+#'
+#' Note that \eqn{ff_i=\sqrt(u_i^2 + v_i^2)}, but the same does not hold for the
+#' height-integrated speeds, i.e. \eqn{ff != \sqrt(u^2 + v^2)} as soon as the
+#' ground speed directions vary with altitude.
+#'
 #' }
```
Rafnuss, line 140: **Documentation**
- Add more comments in the code
- Reorder the subsection in the same order than the Available quantities subsection
- Correct the `\alpha `and `\pi`

R/integrate_profile.R
```diff
@@ -172,35 +173,54 @@ integrate_profile.vp <- function(x, alt_min = 0, alt_max = Inf, alpha = NA,
...
   stopifnot(is.na(alpha) || is.numeric(alpha))
 
+  if (alt_max <= alt_min) stop("'alt_min' should be smaller than 'alt_max'")
+
+  # Altitudinal resolution of the bin
   interval <- x$attributes$where$interval
 
-  if (alt_max <= alt_min) stop("'alt_min' should be smaller than 'alt_max'")
+  # Min altitude is the ground level
+  alt_min <- max(alt_min, min(x$attributes$where$height))
```
Rafnuss, line 182: alt_min is now the ground level rather than the lowest bin

R/integrate_profile.R
```diff
@@ -172,35 +173,54 @@ integrate_profile.vp <- function(x, alt_min = 0, alt_max = Inf, alpha = NA,
...
 
-  alt_min <- max(alt_min, min(x$data$height))
+  # Max altitude is the top of the highest bin
   alt_max <- min(alt_max, max(x$data$height) + interval)
-  if (alt_max - alt_min <= interval) stop(paste("selected altitude range (", alt_min, "-", alt_max, " m) should be wider than the width of a single altitude layer (", interval, " m)", sep = ""))
```
Rafnuss, line 181: This check is not necessary. We can compute the integral for any altitude span

R/integrate_profile.R
```diff
@@ -172,35 +173,54 @@ integrate_profile.vp <- function(x, alt_min = 0, alt_max = Inf, alpha = NA,
...
 
-  index <- which(x$data$height >= alt_min & x$data$height < alt_max)
+  # dh is a vector of the height of each bin used for the computation in km
+  # Its value is zeros for bins completly below alt_min or above alt_max, the
+  # interval value for all bin completly above alt_min and below alt_max, and
+  # a ratio of interval for the bins overlaping alt_min and alt_max
+  dh <- pmin(pmin(pmax(x$data$height-alt_min,0),interval),
+             pmin(pmax(alt_max-x$data$height,0),interval)) / 1000
```
Rafnuss, line 192: dh is the key new element in the code.

Example:
```
> alt_min <- 519 #x$attributes$where$height
> alt_max <- 3100
> interval <- 200
> 
> dh <- pmin(pmin(pmax(x$data$height-alt_min,0),interval),
+              pmin(pmax(alt_max-x$data$height,0),interval)) / 1000
> dh
 [1] 0.000 0.000 0.000 0.081 0.200 0.200 0.200 0.200 0.200 0.200 0.200 0.200 0.200 0.200 0.200 0.100 0.000 0.000 0.000 0.000 0.000 0.000 0.000 0.000 0.000
```

R/integrate_profile.R
```diff
@@ -269,44 +289,56 @@ integrate_profile.vpts <- function(x, alt_min = 0, alt_max = Inf,
...
+             * 3.6  * dh, na.rm = TRUE)
+  # Reflectivity Traffic Rate in cm^2/km/h
+  rtr <- colSums(get_quantity(x, "eta") * cosfactor * get_quantity(x, "ff")
+             * 3.6 * dh, na.rm = TRUE)
+
+  # The following quantites are vertically summed by weighting each altitudinal
+  # bin based on their bird reflectivity value (eta) and the height
+  w <- get_quantity(x, "eta") * dh
+  w[is.na(w)] <- 0
+  w <- sweep(w, 2, colSums(w), FUN="/")
```
Rafnuss, line 334: `w` replaces `dens_quantity` by accounting for both the relative quantitfy of bird (`eta`) and the height considered `dh`. 

`sweep` normalized `w` so that the sum of each of its column (i.e. vp) is 1. Its always to simply multiply an sum in l. 336-339.

Rafnuss:
@peterdesmet @adokter could you help me trouble shoot the check failed. Is it due to the figures ? 
I have tried to comment as much as possible my code and the modification, but I don't really know if this is up to your standard and what is the correct procedure... 
I have start working on a test-integrate_profile, but I don't know if I should push it to this branch also. What do you think?

R/integrate_profile.R
```diff
@@ -269,44 +289,56 @@ integrate_profile.vpts <- function(x, alt_min = 0, alt_max = Inf,
...
-  u <- colSums(get_quantity(x, "u")[index, ] * get_quantity(x, "dens")[index, ],
-    na.rm = TRUE
-  ) / dens_quantity
-  v <- colSums(get_quantity(x, "v")[index, ] * get_quantity(x, "dens")[index, ],
-    na.rm = TRUE
-  ) / dens_quantity
-  ff <- colSums(get_quantity(x, "ff")[index, ] * get_quantity(x, "dens")[index, ],
-    na.rm = TRUE
-  ) / dens_quantity
+  height <- colSums( (get_quantity(x, "height") + interval / 2) * w, na.rm = T)
```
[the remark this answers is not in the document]
adokter: More a general comment: please verify that whenever you place the radar at the bottom of a bin, you retrieve identical results with this modified code

adokter (changes requested):

R/integrate_profile.R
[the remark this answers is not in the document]
adokter: The tests not completing is because the changes to the function introduces differences in the plots. If it turns out to be small numerical differences due to the lower bin then we can update the snaps for those images to the new format. You need R 4.1 and vdiffr package installed to review the changes introduced in the plot, on older R versions the image tests are skipped, which is probably why you didn't have problems with `devtools::check()` locally

adokter:
> @peterdesmet @adokter could you help me trouble shoot the check failed. Is it due to the figures ?
> I have tried to comment as much as possible my code and the modification, but I don't really know if this is up to your standard and what is the correct procedure...
> I have start working on a test-integrate_profile, but I don't know if I should push it to this branch also. What do you think?

The tests not completing is because the changes to the function introduces differences in the plots. If it turns out to be small numerical differences due to the lower bin then we can update the snaps for those images to the new format. You need R 4.1 and vdiffr package installed to review the changes introduced in the plot, on older R versions the image tests are skipped, which is probably why you didn't have problems with devtools::check() locally

ok to add test-integrate_profile to this PR

R/integrate_profile.R
Rafnuss: Ok for using `dens `instead of `eta`.
What about `weight_densdh`? I thought it would be good to keep the word "weight", as it is normalized.
Changes pushed in d4ce62d129768eef031e1e581b0443c89883ce01

R/integrate_profile.R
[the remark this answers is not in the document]
Rafnuss: > I don't understand the interval/2 statements here, doesn't that become obsolete once you include integration with sub-bin resolution?
As `get_quantity(x, "height")` is the lowest edge of the elevation bin (0, 200 etc...), adding `interval/2` allows to compute the mean bin elevation. This would already implemented before (i.e., I didn't change anything there).

> I don't see why it should be present in u,v and ff
very true, I think this is a copy/paste error. Corrected in 605db6af951f179ecfe83c922d1c46a9ab443100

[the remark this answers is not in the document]
Rafnuss: > More a general comment: please verify that whenever you place the radar at the bottom of a bin, you retrieve identical results with this modified code

Yes, I just checked that and the result are the same for the vp and vpts example.

Rafnuss:
@peterdesmet , should I also push the changes in `man/integrate_profile.Rd`? or is it generated automatically?

Rafnuss:
Thanks @adokter for the advice about the test failure. I have looked at the test-figure files and better understand how it works. 
Before I make the change on snaps figures. I wanted to get your opinion on the change in the figure and see if you agree with the change in the code. 

My understanding of the changes is the following. With an ground level of 519m, the contribution of the bin 400-600m is roughly divided by 2. As the result, a reduction of ~1/4-1/2 in the vertically integrated profile of density is not unexpected considering that most of the reflectivity occurs in this first bin. 

Old MTR:
image
New MTR
image

Old VID:
image
New VID:
image

OLD U:
image

New U:
image

For reference: 
`plot(example_vpts)`
image

adokter:
Hi @Rafnuss,
Thanks for adding those figures as a reference, I agree it looks reasonable.

A few things that remain to be decided:
1) should we make this the new default behaviour, and/or do we add a user option to toggle between the two behaviors? I'm inkling towards adding a user-option, as the changes can be quite large numerically, and we want to keep backwards compatibility.
2) how do we deal with antenna height vs ground-level height. The altitude stored is usually antenna height, which is often above ground-level height, because the radar is often placed on a tower.
3) Certain radars scan at negative elevation angles (often in mountainous areas), and therefore do detect birds below the antenna height. The assumption that we do not have data below the radar altitude is therefore not true in general.

Rafnuss:
Thanks @adokter, good points. I agree that this is far from being the ideal solution. And therethere, I think it would make sense to have a parameter to keep this solution evolving while allowing back-compatibility. I see two problems:

**Ground level elevation**
Determining the ground level altitude can become quite complex because of the rough topography of mounatnous area and the fact that radar's locations are baised toward local peaks. Simply using the ground elevation at the radar's location is, in my opinion, not sufficient. Until now, my favorite option has been to look at the average altitude across the 25-35km radius around the weather radar. In addition, to make this a bit more realistic, I only look at maximum altitude projected along the direction of migration. With this later option, we assume that bird do little or no change of altitude accros the 25-35km radius and thus are already flying at the height necessary to cross any topography. This is relatively simple to compute, but I'm not sure how best to implement in bioRad. 

**No data != no bird**
Independently of the elevation used, if no data are available, it will be assume that no bird are present (e.g. `na.rm = TRUE` https://github.com/adokter/bioRad/pull/472/files/740f84a03085a6d8260868b87c92dcc0ff9cf2c7#diff-bb49ee46f3928e72eca331bca0385e3626b5aebd22b36bac7a0296d965934630R195). This is only a problem for the lowest bins overlaping with the elevation used. 
As a simple solution, we could assume the same density, speed, etc.. as the first bin with available data above.

Rafnuss:
- Revert default behaviour to the original (integration throughout the entire vertical profile, intedpendantly of if data is present and the height of the antenna). 
- Add  an option for `alt_min` to be `"antenna"`, in which case, `alt_min` is set to the heigt of the antenna (`x$attributes$where$height`)
- Overall, the main improvement of this PR is the ability to integreate from any alt_min and alt_max without rounding it to the nearest bin. The integration is thus accounting for any volume considered in each bin.

adokter (approved):

adokter (approved):

PR #475, opened 2021-06-30
Update regularize_vpts.R

Rafnuss:
This PR adds functionality to specify the time-difference over which temporal nearest-neighbor interpolation should be applied to regularized timesteps. The argument `fill` is updated to accept a time-interval (specified in units `units`).

- Allow `fill` to be numeric, specifying the time-interval for newly introduced profiles by the regularization to be filled by nearest neighbor interpolation
- Default `fill` setting is changed from `FALSE` to `TRUE`, however this PR introduces no breaking change compared to the previous default (`fill=FALSE`), but it does for the previously non-default setting (`fill=TRUE`). The default fill setting is changed in this PR to `fill=TRUE`, which is mapped to `fill=interval` to maintain backwards compatibility, i.e. the new default `fill=TRUE` is identical to the old default `fill=FALSE`.
-  A message is given when users specify `fill=TRUE` to warn for the updated behavior
- The old behavior `fill=TRUE` can be reproduced using `fill=Inf`
-  Documentation updates of `fill` argument.


Rafnuss:
I just add a closer look at the change in the `plot(example.vpts)`. 

OLD:
image

NEW:
image


There is a lot of empty vertical profile with the suggested changes. This is due to decreasing the maximum windows of the nearest neighbor interpolation. It results that more vertical profile have no match in the original dataset (in this case, more than +/- 2.5 minutes from the new grid).  Here is a close-up on one night to see more details: 

OLD:
image
NEW:
image

The old regularized was basically duplicating the orignal data at two consequentive time step. I think that this is not necessary optimal for later analysis as it would create correlation in the dataset. 

Here is a close-up on a couple of hours of the datetime of the original data (black dot) and the regularlized grid (red cross). I have circled in green new grid node in which I believe an empty vertical profile is more appropriate than the closest neighbour. The blue line indicate the neigherest neighbor interpolation that I would judge correct to do. The yellow circle indicates an original datapoint that is lost in the process (both in the old and new implementation). 
image

Happy to discuss more about that or to know your opinion.

Rafnuss:
Modification of the fill parameters to be a numeric value which define the window search of the nearest neighbor interpolation. `fill` is in unit of interval, so that the windows search is  `fill * interval` (i.e., `+/- fill * interval / 2` .
Fill can also be a logical:

- If `fill==TRUE`, `fill=2`. Original behaviour (also default value)
- If `fill== FALSE`, `fill=0`. (no interpolation).

adokter (approved):

Rafnuss:
Just run into an issue with setting the wrong timezone in `date_min` and `date_max`. I think the code is working as expected (accounting for the difference in timezone. But I was wandering if we should issue a warning message when that's the case (i.e. vpts timezone different from `date_min` and/or `date_max`. 
I can quickly implement it if you think it's useful or maybe it's just me doing that sort of mistake...

PR #485, opened 2021-12-14
add height quantile option to integrate_profile()

adokter:


adokter:
Hi @peterdesmet, no need to review this if you are short on time

I keep running into issues with the snaps in _snaps - my machine wants to update them, but I have to revert to the old ones in order to pass the unit tests on github. Any idea why this happens?

PR #497, opened 2022-02-14
Calculate vp local

bart1:
In this branch I want to do a few thing to make calculating vp's more easy.

- [x] If a local binary of vol2bird is used and no mistnet there is no need to check if the mistnet path has been set
- [x] Since there is a write pvol function how this can work with pvols, directly eliminating the need for each user to write pvols them self. This required to change the way the `mount` argument is implemented but it still behaves the same
- [x] Make sure it is possible to use this function with parallelization like `mclapply` without having conflicts for the file on disk.
The last feature still needs some checking. Easiest would be to ensure the opts file is written to a random file instead of the same file name. This would however change the default behaviour. Any suggestions?


bart1:
> 
> That could work. We would have to change vol2bird from reading a [hard-coded name](https://github.com/adokter/vol2bird/blob/d8365d249fb4a82907a30a7f62e84bfc9a86cc86/lib/constants.h#L167) for the configuration file to making it an input file that the user can specify when calling vol2bird. Issue added at [adokter/vol2bird#186](https://github.com/adokter/vol2bird/issues/186)

Hmm that is quite a long term project. Maybe we should work to that in the long term and for the moment people should set their mount directory to  a unique place? Maybe we can adjust `calculate_vp` so that if the mount dir does not exist it will make and delete it? Then for users it is relatively easy to ensure there is a unique mounting directory. Alternatively we leave it completely to the users and just document that it needs to be done.

Do you have a preference?

adokter:
Hi @bart1, I think of mounting directory as specific to Docker, and since we are working on getting rid of Docker I think that argument will disappear at some point.

I was thinking of an implementation where we could have the user specify (preferably once after installation of the package) the location of the `vol2bird.sh` script as installed by `vol2birdinstall`, and that bioRad would then work with that installation path. The user would then no longer have to specify it in the various commands that currently have a `local_install` argument.

adokter:
FYI there is one other route available: if you set the environment variable `OPTIONS_FILE` in your shell, vol2bird will use the file specified in that environment variable. So that could be another way of having bioRad specify the location of the config file, by having it set the environment variable. But that probably wouldn't work very nicely with parallelizing

bart1:
> Hi @bart1, I think of mounting directory as specific to Docker, and since we are working on getting rid of Docker I think that argument will disappear at some point.
> 
> I was thinking of an implementation where we could have the user specify (preferably once after installation of the package) the location of the `vol2bird.sh` script as installed by `vol2birdinstall`, and that bioRad would then work with that installation path. The user would then no longer have to specify it in the various commands that currently have a `local_install` argument.

One solution I frequently take is to create an option with such a path people can set such an option in `.Rprofile` the package can try some sensible defaults if its not set and it provides the user with flexiblity if different versions are installed. You could then still have it as an argument function: `local_install=options("bioRad.vol2bird_path")`. If the value of that option is `NULL` you can still look in the `PATH`, and some default locations like `opt`

bart1:
This latest push now with a random temporary file, it does hover require an up to date version of vol2bird which is checked for

bart1:
Thanks for the quick fix to vol2bird!

adokter:
@bart1 could you resolve conflicts in this branch, is this one ready for merging or do you still need to run some tests?

bart1:
> @bart1 could you resolve conflicts in this branch, is this one ready for merging or do you still need to run some tests?

Is this branch still relevant? It does some improvement for calculate vp but I'm not sure when the big overhaul of the vol2bird is going to kick in and if this is still relevant. I just resolved the merge conflict it was rather small. If I remember it is backward compatible (besides random location of opts file) so merging it would not hurt I guess.

bart1:
Thx

adokter:
This PR has broken support for input of multiple scans when using a local vol2bird install. Will be fixed in #506

Issue #498, opened 2022-02-15
`write_pvolfile` fails if no conversion object is present


bart1:
I noticed `write_pvolfile` fails if there is no conversion attribute, would it make sense to guess some sensible defaults? Example below. 

``` r
require(bioRad)
#> Loading required package: bioRad
#> Welcome to bioRad version 0.5.2.9494
#> Docker daemon running, Docker functionality enabled (vol2bird version 0.5.0)
pvolfile <- system.file("extdata", "volume.h5", package = "bioRad")
example_pvol <- read_pvolfile(pvolfile)
write_pvolfile(example_pvol, tempfile())
aa<-calculate_param(example_pvol,aa=c(DBZH))
write_pvolfile(aa, tempfile())
#> Error in x[list] <- values: replacement has length zero

aa$scans[[1]]$params$DBZH |> str()
#>  'param' num [1:480, 1:360] NaN NaN NaN NaN -7 -8 -5 1 -9 -13 ...
#>  - attr(*, "radar")= chr "SE50"
#>  - attr(*, "datetime")= POSIXct[1:1], format: "2015-10-18 18:00:00"
#>  - attr(*, "geo")=List of 7
#>   ..$ lat    : num 56.4
#>   ..$ lon    : num 12.9
#>   ..$ height : num 209
#>   ..$ elangle: num 0.5
#>   ..$ rscale : num 500
#>   ..$ ascale : num 1
#>   ..$ rstart : num 0
#>  - attr(*, "param")= chr "DBZH"
#>  - attr(*, "conversion")=List of 5
#>   ..$ gain    : num 1
#>   ..$ offset  : num -31
#>   ..$ nodata  : num 255
#>   ..$ undetect: num 0
#>   ..$ dtype   : chr "H5T_STD_U8BE"
aa$scans[[1]]$params$aa |> str()
#>  'param' num [1:480, 1:360] NaN NaN NaN NaN -7 -8 -5 1 -9 -13 ...
#>  - attr(*, "param")= chr "aa"
#>  - attr(*, "radar")= chr "SE50"
#>  - attr(*, "datetime")= POSIXct[1:1], format: "2015-10-18 18:00:00"
#>  - attr(*, "geo")=List of 7
#>   ..$ lat    : num 56.4
#>   ..$ lon    : num 12.9
#>   ..$ height : num 209
#>   ..$ elangle: num 0.5
#>   ..$ rscale : num 500
#>   ..$ ascale : num 1
[...]

PR #510, opened 2022-05-04
Fix for #498 Infer conversion details if missing

barthoekstra:
If `calculate_param()` causes conversion information to be missing, this will figure out how to write the data (see #498). If all values are  integers and can be stored as 8- or 16-bit integers (including a nodata and undetect value), it will be stored as such. If not, it will simply be stored directly as a 32-bit float, which should be more than sufficient.

@bart1 I think this should be an OK fix?


adokter:
thanks @barthoekstra, merging into the bioRad 0.6 release

PR #517, opened 2022-07-22
integrate vol2birdR into bioRad

adokter:
* Deprecates the use of Docker
* adds vol2birdR as a dependency
* deprecates `local_install` argument in `calculate_vp()` and `apply_mistnet()`


adokter:
Hi @peterdesmet, here a bioRad version using vol2birdR as a dependency. CRAN is taking a summer break, so vol2birdR is currently a remote dependency (see DESCRIPTION), otherwise it is ready for a CRAN submission. I have adjusted for bioRad the `R-CMD-check.yaml` workflow from vol2birdR, making edits in the order and changing `pak::local_install_dev_deps(upgrade = TRUE)` to `pak::local_install_deps(upgrade = FALSE)`, such that the checks also work for OSX.

.github/workflows/R-CMD-check.yaml
```diff
@@ -18,29 +18,48 @@ jobs:
...
+        if: runner.os == 'macOS'
+        run: |
+          env HOMEBREW_NO_AUTO_UPDATE=1 brew install hdf5 proj gsl
+     
+      - uses: r-lib/actions/setup-r-dependencies@v2
+        with:
+          cache-version: 2
+          extra-packages: any::rcmdcheck
+          
       - uses: r-lib/actions/setup-pandoc@v1
```
peterdesmet, line 44: @adokter can this line move above `- uses: r-lib/actions/setup-r@v2` as it originally was? I think it reads easier.

.github/workflows/R-CMD-check.yaml
```diff
@@ -18,29 +18,48 @@ jobs:
...
       matrix:
         config:
-          # - {os: macOS-latest,   r: 'release'}
-          - {os: windows-latest, r: 'release'}
-          - {os: ubuntu-latest,   r: 'devel', http-user-agent: 'release'}
-          - {os: ubuntu-latest,   r: 'release'}
-          - {os: ubuntu-latest,   r: 'oldrel-1'}
-
+          - {os: macOS-latest,   r: 'release'}
+          #- {os: windows-latest, r: 'release'}
```
peterdesmet, line 22: @adokter why are `windows-latest` and `ubuntu-latest:release` disabled (and `ubuntu-latest:oldrel-1` removed)?

Is there a plan to re-enable these? If so, what is the blocker? CRAN release of vol2birdR?

.github/workflows/R-CMD-check.yaml
```diff
@@ -18,29 +18,48 @@ jobs:
...
+          extra-packages: any::rcmdcheck
+          
       - uses: r-lib/actions/setup-pandoc@v1
 
-      - uses: r-lib/actions/setup-r@v1
-        with:
-          r-version: ${{ matrix.config.r }}
-          http-user-agent: ${{ matrix.config.http-user-agent }}
-          use-public-rspm: true
+      - name: Install system dependencies
```
peterdesmet, line 46: @adokter I assume it makes sense to rename this to "Install Linux system dependencies"?

.github/workflows/R-CMD-check.yaml
```diff
@@ -18,29 +18,48 @@ jobs:
...
       - uses: actions/checkout@v2
+      
+      - uses: r-lib/actions/setup-r@v2
 
+      - name: Install macOS system dependencies
+        if: runner.os == 'macOS'
+        run: |
+          env HOMEBREW_NO_AUTO_UPDATE=1 brew install hdf5 proj gsl
+     
+      - uses: r-lib/actions/setup-r-dependencies@v2
```
peterdesmet, line 39: Can this block be moved after "Install Linux system dependencies", so that the whole flow is:

- checkout@v2
- setup-r@v2
- setup-pandoc@v1
- Install MacOS system dep
- Install Linux system dep
- setup-r-dependencies@v2
- Install dependencies
- check-r-package@v2

peterdesmet (commented):
Great that the workflow works! 2 remarks:

- If possible, I would change to order of doing things to make it more logical/readable
- Please clarify why some OS are disabled (is it temporary?)

.github/workflows/R-CMD-check.yaml
adokter: * I think the ubuntu releases can be restored, I just wasn't sure if they were adding anything
* windows-latest blocker is indeed the CRAN release of vol2birdR, we need to have the binary available

.github/workflows/R-CMD-check.yaml
adokter: yes

.github/workflows/R-CMD-check.yaml
adokter: I ran into two issues earlier:
- The Mac systems dependencies need to be installed _before_ calling setup-r-dependencies@v2, since vol2birdR is a dependencies and requires those system libraries.
- The install Linux system dep block uses package `pak`, which is install with setup-r-dependencies@v2 and has to come after that.

So in your above list you have to swap "Install Linux system dep" and "setup-r-dependencies@v2".

.github/workflows/R-CMD-check.yaml
adokter: I reopened this branch so you can make changes as you see fit

.github/workflows/R-CMD-check.yaml
adokter: I'm not sure if setup-pandoc@v1 can be moved up

adokter:
@peterdesmet I reopened https://github.com/adokter/bioRad/tree/vol2birdR to address your review.

.github/workflows/R-CMD-check.yaml
adokter: moving pandoc up works, resolving this one

.github/workflows/R-CMD-check.yaml
adokter: added those other ubuntu releases

PR #536, opened 2022-11-07
csv support for read_vpts()

adokter:
This is a prototype still under development. Supports csv output obtained by concatenated writes of. csv should have a (non-repeated) header

```
write_csv(as.data.frame(example_vp, geo=F, suntime=F), file = xxx.csv)
```


bart1:
Quick remark, if read speed is a consideration, I have good experiences using `vroom` or `readr` for considerably speeding up reading plain text files

peterdesmet:
Yes, I would also recommend using `read_csv()` under the hood.

PR #565, opened 2023-05-16
make `vol2birdR` suggested

### LINKED ISSUE #557: `vol2birdR` as a suggested package
To reduce the dependency tree it might make sense to move `vol2birdR` to suggested packages. As far as I see it is only important for `calculate_vp`. With check installed it is easy enough to check if it is installed see: https://gitlab.com/bartk/move2/-/blob/main/R/movebank_credentials.R#L29 . It would make installing the package easier. And the checks are quickly done with: https://rlang.r-lib.org/reference/is_installed.html

bart1:
This should resolve #557


bart1:
It seems there are locally still some error but they seem unrelated to this push

R/hooks.R
```diff
@@ -1,9 +1,11 @@
 .onLoad <- function(libname, pkgname) {
   register_all_s3_methods() # dynamically registers non-imported pkgs (tidyverse) # nocov
-  assert_that(is.string(find.package("vol2birdR", verbose=F, quiet=T)),msg="package vol2birdR not found. See https://adokter.github.io/vol2birdR/ for install instructions")
+  #  assert_that(is.string(find.package("vol2birdR", verbose=F, quiet=T)),msg="package vol2birdR not found. See https://adokter.github.io/vol2birdR/ for install instructions")
```
peterdesmet, line 3: Can this be removed or does it need to be commented?

peterdesmet (approved):

peterdesmet:
I would like @adokter and/or @iskandari to review this as well.

R/read_pvolfile.R
```diff
@@ -127,9 +128,9 @@ read_pvolfile_body <- function(file, param = c(
     }
   } else {
     pvol_tmp <- tempfile()
-
+    rlang::check_installed("vol2birdR", "to read `rsl` files")
```
adokter: please replace `rsl` files with "`NEXRAD` or `IRIS` files"

R/read_pvolfile.R
```diff
@@ -127,9 +128,9 @@ read_pvolfile_body <- function(file, param = c(
     }
   } else {
     pvol_tmp <- tempfile()
-
+    rlang::check_installed("vol2birdR", "to read `NEXRAD` files")
```
adokter, line 131: This will only work when the package is on CRAN - currently I get a message
```
> read_pvolfile("~/Desktop/KDIX20201005_020122_V06")
ℹ The package "vol2birdR" is required to read `NEXRAD` files
✖ Would you like to install it?

1: Yes
2: No

Selection: 1
Error in loadNamespace(x) : there is no package called ‘vol2birdR’
In addition: Warning message:
package ‘vol2birdR’ is not available for this version of R

A version of this package for your version of R might be available elsewhere,
see the ideas at
https://cran.r-project.org/doc/manuals/r-patched/R-admin.html#Installing-packages 
```
This message doesn't explain to the user where to obtain vol2birdR

adokter: `rlang::check_installed` has `reason` and `action` arguments, those can potentially be used to give a more informative message, and automatically install from github

adokter (changes requested):
current behavior expects the package to be on CRAN. As long that is not the case, we should have a more informative message, and preferably trigger a devtools::github_install action

R/read_pvolfile.R
adokter: All the tests that require vol2birdR currently fail if vol2birdR is not installed, which would be the default. How do we make sure these tests are run on CRAN, i.e. how do we force installation of vol2birdR on CRAN test servers?

bart1: If `vol2birdR` is released within the next week should we bother fixing this for this short time assuming not many people will work with the bleeding edge package?

bart1: Since we will also have to change it back then in a weeks time?

bart1:
Remarks should now be resolved (actually the devel tests now seems fast now)

adokter:
@bart1 on Linux and Mac, vol2birdR requires certain system requirements to be installed *prior* to installing the package form R. If we give the simple `do you want to install?` option I think many people will proceed that without having the system requirements installed. This is why i added the reference to the vol2birdR install instructions in the `check_installed` message. However the message still tempts people to install without having system dependencies present.

Some options:
1. we could determine what platform people are on, and explicitly instruct them which system dependencies to install first. We need to be able to distinguish between Windows, Mac, Ubuntu and systems supporting yum/rpm, because each has different naming conventions for the same dependencies.
2. for linux and mac remove the option to automatically install altogether, only providing it for windows

adokter (changes requested):
we'll need to warn for required system dependencies for vol2birdR on linux and Mac, and avoid people install without having those present.

bart1:
Sounds like a valid concern, this is off course an issue for more packages relying on  system libraries. The situation is also a bit more complex as more linux precompiled packages come available. I will have to do some research what are common solutions

peterdesmet:
All this checks and the message to return are best wrapped up in a helper function `check_vol2birdR()`

bart1:
I have been thinking a bit more about this. I guess generally system requirements are kind strangely handled in R and it would be hard to completely fix this. There would be two distinct routes of installing that would pose the problem at a different time. Either you would install from source or from a binary (more common now there is the posit repository). For compiling from source you would need to have all `-dev` libs installed. If you go for binaries only the libraries are required. Normally I would say the correct installation of packages is a concern of the package it self and not package depending on it but given the close link between `vol2birdR` and `bioRad` we could see how we can facilitate it (although `vol2birdR` should of course also work on it self as it is now). 

Now for the two installation routes:
- If you install from source you would iteratively encounter the missing dependencies during compilation. I have just tried it with the `rocker/tidyverse` image. It seems `hdf5` and `proj` missing are caught during configuration but missing `gsl` and `bz2` i encountered only during compilation. It would be good for the configure of `vol2bird` to at once warn for all missing dependencies (I dont know how difficult that is), now they are encountered one by one.
- If you install from a binary repository (e.g. when `vol2birdR` is on cran posit might provide this, I'm not sure if they have limits for system dependencies when they do not provide binaries). The installation will proceed without errors. The errors would only occur at run time. I do not have a binary for `vol2birdR` but for `keyring` it looks like this:
```r
> rlang::check_installed("keyring")
ℹ The package "keyring" is required.
✖ Would you like to install it?

1: Yes
2: No

Selection: 1
Installing package into ‘/usr/local/lib/R/site-library’
(as ‘lib’ is unspecified)
also installing the dependencies ‘sodium’, ‘filelock’

trying URL 'https://packagemanager.posit.co/cran/__linux__/jammy/latest/src/contrib/sodium_1.2.1.tar.gz'
Content type 'binary/octet-stream' length 500113 bytes (488 KB)
==================================================
downloaded 488 KB

trying URL 'https://packagemanager.posit.co/cran/__linux__/jammy/latest/src/contrib/filelock_1.0.2.tar.gz'
Content type 'binary/octet-stream' length 21806 bytes (21 KB)
==================================================
downloaded 21 KB

trying URL 'https://packagemanager.posit.co/cran/__linux__/jammy/latest/src/contrib/keyring_1.3.1.tar.gz'
Content type 'binary/octet-stream' length 351860 bytes (343 KB)
==================================================
downloaded 343 KB

* installing *binary* package ‘sodium’ ...
* DONE (sodium)
* installing *binary* package ‘filelock’ ...
* DONE (filelock)
* installing *binary* package ‘keyring’ ...
* DONE (keyring)

The downloaded source packages are in
	‘/tmp/RtmpVtYP9k/downloaded_packages’
> require(keyring)
Loading required package: keyring
Error: package or namespace load failed for ‘keyring’ in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '/usr/local/lib/R/site-library/keyring/libs/keyring.so':
  libsecret-1.so.0: cannot open shared object file: No such file or directory
```
I do not think there is a complete solution. Maybe a message like : `For successful installation additionally system dependencies might be required, for details see https://github.com/adokter/vol2birdR#install` attached to the reason could work. We could only issue the message on unix like systems.

bart1 (commented):
Now there should be an informative message about system requirements

bart1:
@adokter does OSX on intel not need the system libraries installed for precompiled binaries? Good to know, that is different from linux

adokter (approved):

adokter:
> @adokter does OSX on intel not need the system libraries installed for precompiled binaries? 

Let me test that before we merge this

adokter:
@bart1 seems to work without having any system libraries installed, except I do get an annoying warning message with the pre-installed binary (which does not disappear when installing the system libraries), see https://github.com/adokter/vol2birdR/issues/50.

Do you know if CRAN has plans to provide M1 mac and linux binaries? I only see Windows and intel mac binaries offered [here](https://cran.r-project.org/web/packages/vol2birdR/index.html)

bart1:
@adokter I'm not sure cran is planning binaries for linux, I don't think so, but I use those at POSIT for fast creation of docker containers. As for OSX I do not really keep up with that.

bart1:
I also directly removed the remote for vol2birdR

PR #570, opened 2023-05-17
Removed rgdal dependency

iskandari:
- Asserts that the output of `wgs_to_pro()j` and `proj_to_wgs()` are SpatialPoints objects and not SpatialPointsDataFrame objects; object type will not be guaranteed by `spTransform` until June 2023 as this is determined by the presence or absence of rgdal

> If `rgdal` is loaded, spTransform() converts the SpatialPoints object to a SpatialPointsDataFrame object, even if we call `sp::spTransform` specifically, R's method dispatch system will still use the `rgdal` version of the function *if* it's available in the namespace and the object being transformed is of a class that `rgdal` provides a method for.   We do not want to assume that `rgdal` is or is not loaded in the namespace, therefore, the output must be asserted. 

- Removed rgdal from NAMESPACE and DESCRIPTION files
- Checked that outputs match expected outputs for function calls of `spTransform` in `project_as_ppi()`
- `rgdal` moves to suggested 
- sets evolution status to 2 on bioRad package load ([will become default in June](https://www.r-bloggers.com/2023/05/sp-evolution-status-examples-of-migration-from-retiring-packages/))
- if evolution status is reset to 0 and rgdal is not installed,  handles error and instructs users to reset with package startup messages `sp::set_evolution_status(2L)`

### DIFF
--- a/tests/testthat/test-data.R
+++ b/tests/testthat/test-data.R
@@ -4,21 +4,15 @@ test_that("Example objects are of the correct class", {
   expect_s3_class(example_vpts, "vpts")
 })
 
-test_that("Example objects are of correct class (2nd test)", {
-  expect_true(is.scan(example_scan))
-  expect_true(is.vp(example_vp))
-  expect_true(is.vpts(example_vpts))
-})
-
 test_that("Example scan includes all five parameters", {
-  expect_true(is.param(get_param(example_scan, 'DBZH')))
-  expect_true(is.param(get_param(example_scan, 'VRADH')))
-  expect_true(is.param(get_param(example_scan, 'RHOHV')))
-  expect_true(is.param(get_param(example_scan, 'ZDR')))
-  expect_true(is.param(get_param(example_scan, 'PHIDP')))
+  expect_no_error(get_param(example_scan, 'DBZH'))
+  expect_no_error(get_param(example_scan, 'VRADH'))
+  expect_no_error(get_param(example_scan, 'RHOHV'))
+  expect_no_error(get_param(example_scan, 'ZDR'))
+  expect_no_error(get_param(example_scan, 'PHIDP'))
 })
 
-test_that("Plotting example scan produces no error", {
+test_that("Example scan can be plotted correctly", {
   expect_no_error(plot(example_scan, param = "DBZH"))
   expect_no_error(plot(example_scan, param = "VRADH"))
   expect_no_error(plot(example_scan, param = "RHOHV"))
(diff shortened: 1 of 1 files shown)


R/project_as_ppi.R
```diff
@@ -151,6 +153,7 @@ sample_polar <- function(param, grid_size, range_max, project, ylim, xlim, k = 4
       gridTopo <- as(grid_size, "SpatialPoints")
     }
   } else {
+    browser()
```
bart1, line 156: I guess this should still be removed

adokter:
@iskandari unit tests require some updates, `test-integrate_to_ppi` and `test-project_as_ppi` use spTransform, which requires rgdal. spTransform should be replaced with st_transform from sf package

bart1:
`withr` is nice for doing this kind of thing: https://withr.r-lib.org/reference/with_envvar.html

bart1:
As `withr` is a dependency of `ggplot2` so we could add it as a temporary dependency until this is resolved, then something like this should work:
```r
withr::withr_envvar(list(sp_evolution_status=2), spTransfrom(....))
```

bart1:
Nice idea to make a function out of it I would only suggest to not export it to prevent having to keep it around long term

iskandari:
> Nice idea to make a function (`spTransform2`) out of it I would only suggest to not export it to prevent having to keep it around long term

Yes that's the approach I took, thanks for the suggestion @bart1 !

bart1:
I see the evolution status needs to be set before loading sp ( https://github.com/edzer/sp/blob/main/R/AAA.R ), I'm not sure how that works when importing a package, but that might complicate matters why the withr solution does not work

iskandari:
Behavior with regards to sp and rgdal is now the following:

On package load, evolution status is automatically set to 2 and user is notified (this will very soon be the default with sp):

```
> library(bioRad)
Welcome to bioRad version 0.7.0.9581
using vol2birdR version 1.0.0 (MistNet not installed)
Assigned sp_evolution_status to 2
```

bioRad functions that use rgdal-based sp functions run without errors, whether or not rgdal is installed on the system.  Currently, sp business as usual will use rgdal if it is installed **even if it is not loaded**.  That is handled by checking the output of spTransform() and if necessary converting back to SpatialPoints

```
> sp::get_evolution_status()
[1] 2
> project_as_ppi(my_scan)
               Plan position indicator (class ppi)

  parameters:  DBZH VRADH RHOHV ZDR PHIDP 
        dims:  201 x 201 pixels 
```

If evolution status is set back to 0 *and* rgdal is not installed, bioRad functions that use rgdal-based sp functions throw an error that is handled with a message to the user to reset the evolution status parameter:

```
> sp::set_evolution_status(0L)
[1] 0
> project_as_ppi(my_scan)
Error: spTransform failed. Try resetting sp_evolution_status: sp::set_evolution_status(2L)
> sp::set_evolution_status(2L)
[1] 2
> project_as_ppi(my_scan)
               Plan position indicator (class ppi)

  parameters:  DBZH VRADH RHOHV ZDR PHIDP 
        dims:  201 x 201 pixels
```

R/hooks.R
```diff
@@ -14,4 +14,9 @@
     packageStartupMessage("Assigning sp_evolution_status to 2. See sp::get_evolution_status()")
     packageStartupMessage("This is required until the 'sp' package deprecates 'rgdal'")
   }
+  if (requireNamespace("sp", quietly = TRUE)) {
```
adokter, line 17: this statement is already present in the 4 lines above it

adokter: I noticed in the readme you still ask users to set the sp_evolution_status manually as well, is that still necessary or does this statement take care of it already?

adokter (changes requested):

Issue #568, opened 2023-05-16
VPTS objects as data.frame


peterdesmet:
This issue summarizes the outcomes of a discussion of May 16, 2023.

## Goal

Get rid of the bioRad objects `vp` and `vpts`. Instead, make them **VPTS data frames**, so they can are compatible with functions outside bioRad, like `dplyr`.

## Design decisions

1. VPTS are tibble dataframes, with an extra class attribute `vpts`:
    ```
    > class(vpts)
    [1] "vpts"        "spec_tbl_df" "tbl_df"      "tbl"         "data.frame"
    ```
2. `print(vpts)` shows summary information and the default tibble preview. The summary information is for the whole data frame, not per radar. Users can reduce that info by running `filter(radar == "bejab")` beforehand:
    ```
    # Irregular time series of vertical profiles (class vpts)
    # Radars: bejab, bewid
    # Time range: 2023-02-01 00:00:00 / 2023-03-01 00:00:00 (2 days)
    # Height range: 0 / 4000
    # A tibble: 85,700 × 26
       radar datetime            height     u       v       w    ff    dd sd_vvp gap     eta  dens    dbz
       <chr> <dttm>               <dbl> <dbl>   <dbl>   <dbl> <dbl> <dbl>  <dbl> <lgl> <dbl> <dbl>  <dbl>
     1 bejab 2023-02-01 00:00:00      0 NA    NA       NA     NA     NA     4.04 TRUE   99.0  9.00  -5.61
     2 bejab 2023-02-01 00:00:00      0 NA    NA       NA     NA     NA     4.01 TRUE  111.  10.1   -5.11
     3 bejab 2023-02-01 00:00:00      0 NA    NA       NA     NA     NA     3.44 TRUE  120.  10.9   -4.79
     4 bejab 2023-02-01 00:00:00    200  4.50 -0.502   66.3    4.53  96.4   2.82 FALSE  58.0  5.27  -7.93
     5 bejab 2023-02-01 00:00:00    200  2.73 -2.19    47.7    3.50 129.    2.58 FALSE  61.9  5.63  -7.65
     6 bejab 2023-02-01 00:00:00    200  1.13 -1.77   -44.4    2.10 147.    3.40 FALSE  68.2  6.20  -7.23
     7 bejab 2023-02-01 00:00:00    400  1.58  0.373   -4.74   1.62  76.7   2.43 FALSE  26.6  2.42 -11.3 
[...]

PR #574, opened 2023-05-17
replace data.table in `regularize_vpts`

bart1:
The goal for this pull request is to reduce dependencies, currently both `data.table` and `dplyr` are there. With @peterdesmet we decided to go for only `dplyr`, as it what we will use for #568.


bart1:
The change to `regularize_vpts` required replacing a roll statement  that replaced a previous base R statement. The roll statement sped up things this solution is slightly slower but its milliseconds work, and I think it should scale better then the previous apply function:
``` r
devtools::load_all("~/bioRad_new/bioRad/")
#> ℹ Loading bioRad
#> Welcome to bioRad version 0.7.0.9547
#> 
#> using vol2birdR version 0.2.2.9043 (MistNet not installed)
ts<-example_vpts
daterange<-lubridate::floor_date(range(ts$datetime), '5 mins')
grid <- seq(from = daterange[1], to = daterange[2], by = 300)


require(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#> The following objects are masked from 'package:bioRad':
#> 
#>     between, filter, first, lag, last
#> 
#> The following object is masked from 'package:testthat':
#> 
#>     matches
#> 
#> 
#>     filter, lag
#> 
#> 
bench::mark(
  datatable=data.table::setDT(data.frame(datetime=ts$datetime))[data.frame(grid), roll = "nearest", which = TRUE, on = "datetime==grid"],
  base=sapply(
    grid,
    function(x) {
      which.min(abs(ts$datetime - x))
    }
  ),
dplyr={  data.frame(grid=grid) %>% 
    left_join(data.frame(ts=ts$datetime, id=seq_along(ts$datetime)),by =join_by(closest(grid>=ts)) ) %>%
    left_join(data.frame(ts=ts$datetime, id=seq_along(ts$datetime)),by =join_by(closest(grid<=ts)),suffix = c('','.after') ) %>%
    mutate(index=if_else((grid-ts.after)>(ts-grid),id.after,id), index=if_else(is.na(index), if_else(is.na(id), id.after,id),index)) %>% pull(index)}
) |> plot()
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.
```



<sup>Created on 2023-05-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/regularize_vpts.R
```diff
@@ -1,3 +1,5 @@
+globalVariables(c("x","y","closestt"))
```
adokter, line 1: can this be removed?

bart1: As far as I can see it it not easy, it might be possible to do something with `rlang::sym` but the code would become very unreadable. It needed for the line where `closest` is used (I see i made a typo here). For other dplyr stuff generally `.data` works but here its not really possible as far as I can see and they are also not exported from `dplyr`

bart1: The note here is a result of this typo:
https://github.com/adokter/bioRad/actions/runs/5003819799/jobs/8965543985#step:9:96

adokter (approved):

Issue #376, opened 2020-04-30
reduce package size to below 5MB


adokter:
Addition of unit tests has brought the package size (again) above 5MB:

Tarball now 5012207 bytes, need to shed at least 12207 bytes for a CRAN submission

```
$ du -hc
476K	./man/figures
828K	./man
588K	./vignettes
204K	./tests/testthat
208K	./tests
400K	./R
4.0K	./build
1.7M	./inst/extdata
1.3M	./inst/doc
2.9M	./inst
1.4M	./data
6.3M	.
6.3M	total
```

adokter:
@peterdesmet we either have to tackle #376 within the next week or so, but I think it will be quite a bit of work to do that properly

My suggestion is to reduce the size of https://github.com/adokter/bioRad/blob/master/vignettes/ecog12354-fig-0002-m.jpg which you added as part of the bioRad vignette

peterdesmet:
I can add the external files to Zenodo quite easily, but I guess updating all the examples will be some work?

Can’t we hotlink the images to the url of the paper for now?

adokter:
> I guess updating all the examples will be some work?

Yes, and so many tests/examples depend on them that it slows down checks quite a bit, so prefer waiting

> Can’t we hotlink the images to the url of the paper for now?

Maybe, worth a try - what we need is the figure not showing up (or being reduced in size) in the tarball that is generated when you run `R CMD build bioRad`. Not sure if the file stays out of the tarball during vignette generation, even when you hotlink.

peterdesmet:
Hotlinked in e0b3f3b, but the figure does get encoded in the bioRad.html in the tarball 😕. Total size of tarball 4.5MB.

adokter:
```
du -hc man vignettes tests R inst data
528K    man/figures
908K    man
 68K    vignettes
260K    tests/testthat
264K    tests
464K    R
2.6M    inst/extdata
2.6M    inst
1.4M    data
5.7M    total
```

adokter:
@bart1 it's mostly the radar example data we need for running the examples

bart1:
I just checked gz compressing vpts.txt does not help

peterdesmet:
There is the suggestion to store the raw example data on Zenodo: https://github.com/adokter/bioRad/issues/317. Doing so might need updates to the examples (`{dontrun}`) and tests (`skip_if_offline()`).

adokter:
[...]

PR #577, opened 2023-05-17
Unit tests plots

CeciliaNilsson709:
Add the snapshots back to unit tests of the plotting functions, and ignore created images in the package build (thanks to @bart1!)


adokter:
Thanks @CeciliaNilsson709 - we also need to make sure none of the snap reference images get included in a CRAN build that we upload, because that will make the package larger than 5 Mb (https://github.com/adokter/bioRad/issues/376). I think adding relevant directories to .Rbuildignore should be able to accomplish that

CeciliaNilsson709:
That is already included in the pull request, thanks to Bart :)

bart1:
I can now reproduce the error local I just have to figure out what is really the problem

bart1:
It seems these test also fail on my computer outside of `covr`, therefore I'm not sure if this is actually the cause, that I'm reproducing. I guess to be sure we would need to adjust the workflows to get the artifacts from github.

For me the differences between the plots from @CeciliaNilsson709 and me are around the diagonal in the ppi plots and the date axis in some other plots. I have attached the `_snaps` directory for reference. If you unzip this and call: `testthat::snapshot_review(path='bioRad-tests/testthat/')` (replacing the path with the path to the directory that contains the unzipped `_snaps` directory) you will be able to visualize the differences. 

- The capitalization of the dates maybe relates to the difference between OS and time libraries, because one is capitalized and one not. 
- I'm less sure where the differences in the diagonal of the ppi plot come from

[_snaps.zip](https://github.com/adokter/bioRad/files/11511945/_snaps.zip)

PS I noticed mainly the vpts plots are somewhat big in file size of the uncompressed svg. One option would be to only plot those for half a day or so instead of the full example vpts.

PS2 I now see why those tests are not failing in the ubuntu and MACOS test in the github action. Because they apparently are based on the build package the reference plots are not in there (see `.Rbuildignore`) and thus new images are created instead of compared e.g.: https://github.com/adokter/bioRad/actions/runs/5003646884/jobs/8965139469?pr=577#step:9:827

bart1:
Did someone on windows run these test and see if the snaps are the same?

CeciliaNilsson709:
It sounds likely from your tests then that the problem probably comes back to slight differences between different systems then right?

I tried to look around a bit but mostly found others with similar problems and not really any solutions. One way to move forward quickly could be to skip these tests in GitHub actions, `skip_on_ci()`, and just use these tests locally. I guess it mostly depends on what we think the purpose of these tests are? Would it make sense to use them just as local tests when changing things (and then remember to start by creating new snaps?) or does that make it a bit pointless?

bart1:
> It sounds likely from your tests then that the problem probably comes back to slight differences between different systems then right?
> 
> I tried to look around a bit but mostly found others with similar problems and not really any solutions. One way to move forward quickly could be to skip these tests in GitHub actions, `skip_on_ci()`, and just use these tests locally. I guess it mostly depends on what we think the purpose of these tests are? Would it make sense to use them just as local tests when changing things (and then remember to start by creating new snaps?) or does that make it a bit pointless?

Yes its a bit tricky, maybe in some cases we could add a few arguments to make it more standardized (i.e. fix date range). 

But the problem between different systems will keep on popping up, as were all developing on different systems it might be hard to design tests that are stable enough and still meaning full. It would definitely be good to have them around in case were changing things to test manual. It would just mean we would first have to generate existing plots each time and then implement the changes to see if something changes. This is quite elaborate therefore the online cicd would be nice, especially as plots could be quite sensitive for changes in the underlying functions. Maybe accepting my snaps would make the tests less likely to fail as they are made on linux but better would be to get them from the workflows directly and have them actually as an artifact so they are easier check and update in the future.

CeciliaNilsson709:
> but better would be to get them from the workflows directly and have them actually as an artifact so they are easier check and update in the future.

Are you then suggesting saving artefacts from each systems workflow, or how do you mean? 

As a side note: I looked into whether this would be easier if we changed all plot functions to ggplot (currently only plot.scan uses ggplot while the others use base plot). However the gg objects (not the rendered plots) are quite big, so saving those would take up some space (although I guess these could also be skipped in the build for CRAN?). So far I am also struggling with getting saved ggplot objects to be identical to newly created ones, and then of course if going down this path we would need to start by changing the plotting functions to ggplot. I do seem to remember that we wanted to do this anyway, but so far I am not sure it would help with this issue here.

PietrH:
> As a side note: I looked into whether this would be easier if we changed all plot functions to ggplot (currently only plot.scan uses ggplot while the others use base plot). However the gg objects (not the rendered plots) are quite big, so saving those would take up some space (although I guess these could also be skipped in the build for CRAN?). So far I am also struggling with getting saved ggplot objects to be identical to newly created ones, and then of course if going down this path we would need to start by changing the plotting functions to ggplot. I do seem to remember that we wanted to do this anyway, but so far I am not sure it would help with this issue here.

You can also check ggplot objects for specific properties instead of saving the object as a snapshot. So depending on what it is exactly that you are trying to test, this would work as well. If there are variations on the input data (floating point errors for example), you can use `testthat::expect_equal()` instead of `testthat::expect_identical()`.

For example:

``` r
sample_df <- data.frame(
  group = factor(rep(letters[1:3], each = 10)),
  value = rnorm(30)
)
group_means_df <- setNames(
  aggregate(value ~ group, sample_df, mean),
  c("group", "group_mean")
)

plot_out <- 
  ggplot2::ggplot(data = sample_df) +
  ggplot2::geom_point(mapping = ggplot2::aes(x = group, y = value)) +
  ggplot2::geom_point(
    mapping = ggplot2::aes(x = group, y = group_mean), data = group_means_df,
    colour = 'red', size = 3
  )

dplyr::glimpse(plot_out)
#> List of 9
#>  $ data       :'data.frame': 30 obs. of  2 variables:
#>   ..$ group: Factor w/ 3 levels "a","b","c": 1 1 1 1 1 1 1 1 1 1 ...
#>   ..$ value: num [1:30] 0.967 -0.493 -1.576 -0.907 -0.551 ...
#>  $ layers     :List of 2
#>   ..$ :Classes 'LayerInstance', 'Layer', 'ggproto', 'gg' <ggproto object: Class LayerInstance, Layer, gg>
#>     aes_params: list
#>     compute_aesthetics: function
#>     compute_geom_1: function
#>     compute_geom_2: function
#>     compute_position: function
#>     compute_statistic: function
#>     computed_geom_params: NULL
#>     computed_mapping: NULL
#>     computed_stat_params: NULL
#>     constructor: call
#>     data: waiver
#>     draw_geom: function
#>     finish_statistics: function
#>     geom: <ggproto object: Class GeomPoint, Geom, gg>
#>         aesthetics: function
#>         default_aes: uneval
#>         draw_group: function
#>         draw_key: function
#>         draw_layer: function
#>         draw_panel: function
#>         extra_params: na.rm
#>         handle_na: function
#>         non_missing_aes: size shape colour
#>         optional_aes: 
#>         parameters: function
#>         rename_size: FALSE
#>         required_aes: x y
#>         setup_data: function
#>         setup_params: function
#>         use_defaults: function
#>         super:  <ggproto object: Class Geom, gg>
#>     geom_params: list
#>     inherit.aes: TRUE
#>     layer_data: function
#>     map_statistic: function
#>     mapping: uneval
#>     position: <ggproto object: Class PositionIdentity, Position, gg>
#>         compute_layer: function
#>         compute_panel: function
#>         required_aes: 
#>         setup_data: function
#>         setup_params: function
#>         super:  <ggproto object: Class Position, gg>
#>     print: function
#>     setup_layer: function
#>     show.legend: NA
#>     stat: <ggproto object: Class StatIdentity, Stat, gg>
#>         aesthetics: function
#>         compute_group: function
#>         compute_layer: function
#>         compute_panel: function
#>         default_aes: uneval
#>         dropped_aes: 
#>         extra_params: na.rm
#>         finish_layer: function
#>         non_missing_aes: 
#>         optional_aes: 
#>         parameters: function
#>         required_aes: 
#>         retransform: TRUE
#>         setup_data: function
#>         setup_params: function
#>         super:  <ggproto object: Class Stat, gg>
#>     stat_params: list
#>     super:  <ggproto object: Class Layer, gg> 
#>   ..$ :Classes 'LayerInstance', 'Layer', 'ggproto', 'gg' <ggproto object: Class LayerInstance, Layer, gg>
#>     aes_params: list
#>     compute_aesthetics: function
#>     compute_geom_1: function
#>     compute_geom_2: function
#>     compute_position: function
#>     compute_statistic: function
#>     computed_geom_params: NULL
#>     computed_mapping: NULL
#>     computed_stat_params: NULL
#>     constructor: call
#>     data: data.frame
#>     draw_geom: function
#>     finish_statistics: function
#>     geom: <ggproto object: Class GeomPoint, Geom, gg>
#>         aesthetics: function
#>         default_aes: uneval
#>         draw_group: function
#>         draw_key: function
#>         draw_layer: function
#>         draw_panel: function
#>         extra_params: na.rm
#>         handle_na: function
#>         non_missing_aes: size shape colour
#>         optional_aes: 
#>         parameters: function
#>         rename_size: FALSE
#>         required_aes: x y
#>         setup_data: function
#>         setup_params: function
#>         use_defaults: function
#>         super:  <ggproto object: Class Geom, gg>
#>     geom_params: list
#>     inherit.aes: TRUE
#>     layer_data: function
#>     map_statistic: function
#>     mapping: uneval
#>     position: <ggproto object: Class PositionIdentity, Position, gg>
#>         compute_layer: function
#>         compute_panel: function
#>         required_aes: 
#>         setup_data: function
#>         setup_params: function
#>         super:  <ggproto object: Class Position, gg>
#>     print: function
#>     setup_layer: function
#>     show.legend: NA
#>     stat: <ggproto object: Class StatIdentity, Stat, gg>
#>         aesthetics: function
#>         compute_group: function
#>         compute_layer: function
#>         compute_panel: function
#>         default_aes: uneval
#>         dropped_aes: 
#>         extra_params: na.rm
#>         finish_layer: function
#>         non_missing_aes: 
#>         optional_aes: 
#>         parameters: function
#>         required_aes: 
#>         retransform: TRUE
#>         setup_data: function
#>         setup_params: function
#>         super:  <ggproto object: Class Stat, gg>
#>     stat_params: list
#>     super:  <ggproto object: Class Layer, gg> 
#>  $ scales     :Classes 'ScalesList', 'ggproto', 'gg' <ggproto object: Class ScalesList, gg>
#>     add: function
#>     clone: function
#>     find: function
#>     get_scales: function
#>     has_scale: function
#>     input: function
#>     n: function
#>     non_position_scales: function
#>     scales: list
#>     super:  <ggproto object: Class ScalesList, gg> 
#>  $ mapping    : Named list()
#>   ..- attr(*, "class")= chr "uneval"
#>  $ theme      : list()
#>  $ coordinates:Classes 'CoordCartesian', 'Coord', 'ggproto', 'gg' <ggproto object: Class CoordCartesian, Coord, gg>
#>     aspect: function
#>     backtransform_range: function
#>     clip: on
#>     default: TRUE
#>     distance: function
#>     expand: TRUE
#>     is_free: function
#>     is_linear: function
#>     labels: function
#>     limits: list
#>     modify_scales: function
#>     range: function
#>     render_axis_h: function
#>     render_axis_v: function
#>     render_bg: function
#>     render_fg: function
#>     setup_data: function
#>     setup_layout: function
#>     setup_panel_guides: function
#>     setup_panel_params: function
#>     setup_params: function
#>     train_panel_guides: function
#>     transform: function
#>     super:  <ggproto object: Class CoordCartesian, Coord, gg> 
#>  $ facet      :Classes 'FacetNull', 'Facet', 'ggproto', 'gg' <ggproto object: Class FacetNull, Facet, gg>
#>     compute_layout: function
#>     draw_back: function
#>     draw_front: function
#>     draw_labels: function
#>     draw_panels: function
#>     finish_data: function
#>     init_scales: function
#>     map_data: function
#>     params: list
#>     setup_data: function
#>     setup_params: function
#>     shrink: TRUE
#>     train_scales: function
#>     vars: function
#>     super:  <ggproto object: Class FacetNull, Facet, gg> 
#>  $ plot_env   :<environment: R_GlobalEnv> 
#>  $ labels     :List of 2
#>   ..$ x: chr "group"
#>   ..$ y: chr "value"
#>  - attr(*, "class")= chr [1:2] "gg" "ggplot"

testthat::test_that("labels are on the right axis",{
  testthat::expect_identical(
    plot_out$labels,
    list(x = "group", y = "value")
  )
})
#> Test passed 🎉
```

<sup>Created on 2023-05-24 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

CeciliaNilsson709:
Thanks both! After thinking more about it I took some decisions to move this along. 

TLDR; I don't think the work we would need to put into making the snapshots work is worth it in our case. I therefore removed the snapshot tests and slightly extended the other testing as far as possible with the current plot functions. 


**To document the some of our reasoning and conclusions:**
- Most of the plot functions use base R, and are therefore difficult to test other than by taking snapshots of the output plots. This is however sensitive to small visual changes due to different systems etc, see above discussion. Changing all plotting functions to ggplot offers more flexibility for testing, but is a large undertaking. 

- I don't really see the use case for us in which failing snapshot tests of plots will signify something important. The example files are currently artefacts so changes in underlying data and processing would not be caught though snapshot tests. Changes to the plot functions themselves would presumably change something in the visual output, so failing snapshots would be of limited use, and require manual inspection anyway. And in the case we want to change plot functions but maintain identical visual output that is easier tested locally.

- Plotting visually identical plots is not important functionality for bioRad. Plotting _correct_ plots is however of course important, and if/when we move to ggplot we can more carefully test this.

tests/testthat/test-plot.ppi.R
```diff
@@ -2,29 +2,31 @@ data(example_scan)
 ppi <- project_as_ppi(example_scan)
 example.plot <-plot(ppi, param = "VRADH", zlim = c(-5, 5))
 
-test_that("plot.ppi() returns error on incorrect parameters", {
```
peterdesmet, line 5: In the test title I would retain `plot.ppi()` cf. how you have titled other tests. Affects all tests in this file.

tests/testthat/test-plot.ppi.R
```diff
@@ -2,29 +2,31 @@ data(example_scan)
...
+test_that("plot() returns error on incorrect parameters", {
+  # use plot.ppi() to avoid defaulting to base plot()
   expect_error(plot.ppi("not_a_ppi"))
-  expect_error(plot.ppi(ppi, param = "not_a_param"))
-  expect_error(plot.ppi(ppi, quantity = "DBZH"))
+  expect_error(plot(ppi, param = "not_a_param"))
+  expect_error(plot(ppi, quantity = "DBZH"))
 })
 
 test_that("if param missing default to DBZH", {
```
peterdesmet, line 12:
```suggestion
test_that("plot.ppi() defaults to DBZH on missing param", {
```

tests/testthat/test-plot.ppi.R
```diff
@@ -2,29 +2,31 @@ data(example_scan)
...
-  expect_equal(plot.ppi(ppi), plot.ppi(ppi, param = "DBZH"))
+  expect_equal(plot(ppi), plot(ppi, param = "DBZH"))
 })
 
-test_that("plot.ppi() returns ggplot object", {
-  expect_true(all(class(plot.ppi(ppi)) == c("gg", "ggplot")))
+test_that("plot() returns ggplot object", {
+  expect_true(all(class(plot(ppi)) == c("gg", "ggplot")))
```
peterdesmet, line 17:
```suggestion
  expect_s3_class(plot(ppi), c("gg", "ggplot"))
```

Please use `expect_s3_class()` for class validation

tests/testthat/test-plot.scan.R
```diff
@@ -1,57 +1,38 @@
...
 })
 
 test_that("plot.scan() defaults to DBZH on missing param", {
-  expect_equal(plot.scan(example_scan), plot.scan(example_scan, param = "DBZH"))
+  expect_equal(plot(example_scan), plot(example_scan, param = "DBZH"))
 })
 
 test_that("plot.scan() returns ggplot object", {
-  expect_true(all(class(plot.scan(example_scan)) == c("gg", "ggplot")))
+  expect_true(all(class(plot(example_scan)) == c("gg", "ggplot")))
```
peterdesmet, line 16: Please use `expect_s3_class()`

tests/testthat/test-plot.scan.R
```diff
@@ -1,57 +1,38 @@
...
 })
 
-test_that("plot.scan() takes zlim arguments", {
-  expect_true(all(example.plot$data$VRADH >= -5, na.rm = TRUE))
-  expect_true(all(example.plot$data$VRADH <= 5, na.rm = TRUE))
+test_that("plot.scan() uses the default zlim from color_scale", {
+  expect_equal(range(example.plot$data$VRADH, na.rm = TRUE), get_zlim("VRADH", zlim))
+
```
peterdesmet, line 21: Avoid empty line, also in two next tests

tests/testthat/test-plot.vp.R
```diff
@@ -1,43 +1,21 @@
...
+  # use plot.vp() to avoid defaulting to base plot()
   expect_error(plot.vp("not_a_vp"))
-  expect_error(plot.vp(example_vp, quantity = "not_a_quantity"))
+  expect_error(plot(example_vp, quantity = "not_a_quantity"))
 
   # Test error on "param" instead of "quantity"
-  expect_error(plot.vp(example_vp, param = "ff"))
+  expect_error(plot(example_vp, param = "ff"))
 
   # Test warning for deprecated arguments
```
peterdesmet, line 9: I think these are best bundled in a separate tests, called "plot.vp() warns for deprecated arguments"

tests/testthat/test-plot.vp.R
```diff
@@ -1,43 +1,21 @@
...
-#     plot.vp(example_vp, quantity = "DBZH")
-#   )
-# })
+
+test_that("plot.vp() produces plots", {
+  expect_s3_class(recordPlot(plot(example_vp)), "recordedplot")
+  expect_s3_class(recordPlot(plot(example_vp, quantity = "ff")), "recordedplot")
+  expect_s3_class(recordPlot(plot(example_vp, quantity = "DBZH")), "recordedplot")
+  expect_s3_class(recordPlot(plot(example_vp, quantity = "dens")), "recordedplot")
+
```
peterdesmet, line 20: Empty line

tests/testthat/test-plot.vpi.R
```diff
@@ -1,65 +1,22 @@
...
+  # use plot.vpts() to avoid defaulting to base plot()
   expect_error(plot.vpi("not_a_vpi"))
-  expect_error(plot.vpi(example_vpi, quantity = "not_a_quantity"))
+  expect_error(plot(example_vpi, quantity = "not_a_quantity"))
 
   # Test error on "param" instead of "quantity"
-  expect_error(plot.vpi(example_vpi, param = "ff"))
+  expect_error(plot(example_vpi, param = "ff"))
 
   # Test warning for deprecated arguments
```
peterdesmet, line 11: Please move to separate test called "plot.vpi() warns for deprecated arguments"

tests/testthat/test-plot.vpi.R
```diff
@@ -1,65 +1,22 @@
...
-#     "example_plot_height",
-#     plot.vpi(example_vpi, quantity = "height")
-#   )
-# })
+test_that("plot.vpi() produces plots", {
+  expect_s3_class(recordPlot(plot(example_vpi)), "recordedplot")
+  expect_s3_class(recordPlot(plot(example_vpi, quantity = "vir")), "recordedplot")
+  expect_s3_class(recordPlot(plot(example_vpi, quantity = "mtr")), "recordedplot")
+  expect_s3_class(recordPlot(plot(example_vpi, quantity = "dd")), "recordedplot")
+
```
peterdesmet, line 21: Empty line

tests/testthat/test-plot.vpts.R
```diff
@@ -1,66 +1,40 @@
...
-#     "example_plot_v",
-#     plot.vpts(example_vpts_reg,quantity = "v")
-#   )
-#
-# })
+test_that("plot.vpts() produces plots", {
+  expect_s3_class(recordPlot(plot(example_vpts_reg)), "recordedplot")
+  expect_s3_class(recordPlot(plot(example_vpts_reg, quantity = "eta")), "recordedplot")
+  expect_s3_class(recordPlot(plot(example_vpts_reg, quantity = "DBZH")), "recordedplot")
+
```
peterdesmet, line 39: Empty line

peterdesmet (changes requested):
Completely agree that the snapshots are not worth it and will cause more trouble than gain. Best to invest our efforts in ggplot.

Made some minor suggestions to style of tests.

iskandari (approved):

iskandari (approved):

peterdesmet (approved):

Issue #553, opened 2023-04-04
Create `download_vpts_aloft()` function


peterdesmet:
```R
download_vpts_aloft(
  date_min = NULL,
  date_max = NULL,
  radars = NULL,
  directory = ".",
  overwrite = FALSE,
  format = "csv" # also hdf5
  source = "baltrad", # also ecog-04003
)
```

- [x] Decide on controlled values for `format`. Either `csv`/`hdf5` (widely applicable) or `monthly`, `daily`, `hdf5`
- [x] Should we make use of an S3 dependency to download files?
- [x] How should we warn for unrecognized radars?
- [x] How should we warn for dates without data?

## Todo

- [ ] Update bucket endpoint from `https://lw-enram.s3-eu-west-1.amazonaws.com` to `s3://aloft`
- [ ] Adapt to new directory structure at bucket
- [x] Add new parameter `source = BALTRAD`
- [x] `radars` parameter will remain the same (5 letter code: `bejab`)
- [ ] `directory` parameter can remain the same
- [ ] `overwrite` parameter can remain the same
- [ ] Update documentation
- [ ] Update tests
- [ ] Deprecate `download_vpfiles()` to `download_vpts_aloft(format = "hdf5")`

PietrH:
I'd like to work on this, here are my thoughts: 

> Decide on controlled values for format. Either csv/hdf5 (widely applicable) or monthly, daily, hdf5
I prefer csv/hdf5, **and return in the same directory structure as the bucket**
> Should we make use of an S3 dependency to download files?
Let's try to do it without the dependency at first and reevaluate.
> How should we warn for unrecognized radars?
`radars: {missing_radars} not found, found {number_of_radars} other radars`
> How should we warn for dates without data?
`No radars found for all dates, radars found for {first_date_found} to {last_date_found}`

## Tests

- Can we use the existing tests from `download_vpfiles()`?
- Testing the directory structure
- Should we test the output files themselves using snapshots? 
- test the new warnings
- test overwriting
- test format


## Branch

Has work already started on this? Is there a branch I can continue on?

PietrH:
If a radar is missing, stop.

PietrH:
- A json file exists with a list of radars: https://github.com/enram/aloftdata.eu/blob/main/_data/OPERA_RADARS_DB.json
- allow downloading multiple radars 
- download_vpfiles() as an example
- error if radar doesn't exist based on json file
[...]

PR #579, opened 2023-05-17
list_vpts_aloft()

PietrH:
# STATUS: READY FOR REVIEW

> https://github.com/adokter/bioRad/issues/553#issuecomment-1551132164

We have decided to build a function list_vpts_aloft() that returns a vector of urls that are known to exist, given the filtering parameters originally envisioned for download_vpts_alof()

```R
list_vpts_aloft(
  date_min = NULL,
  date_max = NULL,
  radars = NULL,
  # directory = ".", This parameter is removed
  # overwrite = FALSE, This parameter is removed
  format = "csv" # also hdf5
  source = "baltrad", # also ecog-04003
)
```

Checking if a file exists can be done using the aws.s3 dependency via: `aws.s3::get_bucket_df(bucket = "s3://aloft", prefix="baltrad/monthly", region = "eu-west-1", max = 2000)` or much slower using `httr: urls[!furrr::future_map_lgl(urls, ~httr::http_error(httr::HEAD(.x)))]`

---

## Dependencies

- I'm using `dplyr` and `magrittr` pipes, so depend on #574 to introduce them into the namespace.

## Scope

Currently only `format = "csv"` is supported

## Todo

- [x] unit tests
- [x] roxygen documentation
- [x] handle missing dates: return all data
- [x] bump version number
- [x] `aws.s3` to Suggests
- [ ] ~`purrr` & `jsonlite` to Suggests~
- [ ] ~consider using lubridate in filter instead of strings~
- [ ] ~consider removing hard coded dates when not provided~
- [x] fix pkgdown bug: `_pkgdown.yml`
- [ ] ~Stop when a radar is missing~
- [x] Add `verbose` argument to silence warnings
- [x] warn twice when both dates and radars are missing
- [x] Update `NEWS.md`
- [x] remove `jsonlite` dependency
- [x] consider `extract_string` helper: used twice
- [x] remove `purrr` dependency
- [x] emtpy vector and warning on no data


PietrH:
Consider moving `jsonlite` and `purrr` to suggests and checking them with `rlang::check_installed()`

PietrH:
Currently the function stops on an error if a radar station is provided that doesn't exist, however, when data for an existing radar station is simply missing, the function will offer a warning and continue to provide links for all other available data. 

Do we want to stop on an error when no data is available for any of the provided radar stations?

peterdesmet:
> when data for an existing radar station is simply missing, the function will offer a warning and continue to provide links for all other available data.

Good. Does it warn only once per radar (preferred) or also when time ranges are missing. I'm also fine with not having a warning for any missing data.

> Do we want to stop on an error when no data is available for any of the provided radar stations?

No, just return an empty vector of paths

PietrH:
I've decided to remove my `purrr::map_chr()` call with a very slightly slower (and more difficult to read) `sapply` to avoid having to state `purrr` as an `Imports`, even tough it's an implicit dependency already trough `tidyr`.

If this is overly defensive, feel free to revert d2c1e6574522634d101b8e0d5bf68f22d8ee73fc. I vastly prefer using `purrr` over base apply functions, but we might get a CRAN warning if we have to many Imports.

PietrH:
> Good. Does it warn only once per radar (preferred) or also when time ranges are missing. I'm also fine with not having a warning for any missing data.

The warnings for radars and time ranges are independent, so it'll warn first if you are requesting a radar station that exists but has no data for your time range. Then it'll warn again if data was found for only a subset of the date range that was requested. 

So you'll get two warnings at the maximum. 

> No, just return an empty vector of paths

Currently we are stopping on the error `No data found for radars between {date_min} - {date_max}`, I'll convert this to a warning and return an empty vector.

PietrH:
Ready for review

iskandari (approved):
Looks good!

PR #582, opened 2023-05-20
Prepend all external functions

### LINKED ISSUE #578: Be explicit about dependency use
Rather than using `import()` and `importFrom()` (currently used in [bioRad.R](https://github.com/adokter/bioRad/blob/master/R/bioRad.R) and functions), use `pkg::function()` notation when using a function from a dependency.

Advantages:
- More welcoming to new maintainers: where is this function coming from?
- Easier to use GitHub/grep find to see dependency use
- No need to maintain `importFrom()`
- Avoids masking/name space clashes

Note: only base R and bioRad function would be written as `function()`. We could also make an exception for pipes (`%>%`).

Example:

```R
# Use
groups <- rhdf5::h5ls(file, recursive = FALSE)$name
# Rather than
groups <- h5ls(file, recursive = FALSE)$name
```

peterdesmet:
Fixes #578.

# Results

- All external functions are prepended with `function::`
- All `import` and `importFrom` lines are removed from the package.

Exceptions:

1. Two dplyr functions are imported, because `dplyr::%>%` and `dplyr::.data` are awkward to write.

https://github.com/adokter/bioRad/blob/561bfc8351fabf8d6fc0cacb593d4bc404235bae/R/bioRad.R#L14

2. In the tests, `testthat::` functions do not need to be prepended.


peterdesmet:
I notice one failure, will investigate later.

adokter (approved):

adokter:
looking good, please bump version before merging

peterdesmet:
I am getting the following NOTE when running R CMD check:

> ❯ checking dependencies in R code ... NOTE
>  Namespace in Imports field not imported from: ‘rgdal’
>   All declared Imports should be used.

This note disappears if I remove `rgdal` from the dependencies. @adokter @bart1 Are we no longer using it?

bart1:
I guess this all should be resolved with #570 @iskandari do you need help there?

peterdesmet:
This PR is now getting failed tests where it didn't before:

```
* checking examples ... ERROR
Error: Running examples in ‘bioRad-Ex.R’ failed
The error most likely occurred in:
> base::assign(".ptime", proc.time(), pos = "CheckExEnv")
> ### Name: calculate_param
```

@adokter any idea what is causing this?

bart1:
@peterdesmet I guess that can have something to do with tidy eval or re export I will investigate

bart1:
Actually it seems to relate to the `rgdal` issue: https://github.com/adokter/bioRad/actions/runs/5037210509/jobs/9045061269?pr=582#step:9:151

That might be closer to resolving here: https://github.com/iskandari/bioRad/pull/1

iskandari:
rgdal issue has been addressed here, passes checks https://github.com/adokter/bioRad/pull/570

R/hooks.R
```diff
@@ -1,9 +1,16 @@
 .onLoad <- function(libname, pkgname) {
   register_all_s3_methods() # dynamically registers non-imported pkgs (tidyverse) # nocov
-  assert_that(is.string(find.package("vol2birdR", verbose=F, quiet=T)),msg="package vol2birdR not found. See https://adokter.github.io/vol2birdR/ for install instructions")
+  assertthat::assert_that(assertthat::is.string(find.package("vol2birdR", verbose=F, quiet=T)),msg="package vol2birdR not found. See https://adokter.github.io/vol2birdR/ for install instructions")
 }
```
iskandari, line 4: I think this will always error out because if the package isn't found a char(0) is returned which doesn't pass the assertion.

```
> assertthat::assert_that(is.string((find.package("foobar", verbose=F, quiet=T))))
Error: (find.package("foobar", verbose = F, quiet = T)) is not a string (a length one character vector).
```

peterdesmet: Do I assume correctly that this PR did not break this line of code but that is was broken already? I leave it to you on how to best fix it. 😄

iskandari: Yes, you're right, it was broken before this PR.   WIll implement the change suggested above.

iskandari (approved):

PR #584, opened 2023-05-22
make `aws.s3` suggested (#583)

### LINKED ISSUE #583: Making `aws.s3` suggested
It would be nice to make `aws.s3` suggested. It adds quite a few dependencies. In total 11 other package need to be installed because of the direct import of `aws.s3`. Currently only `download_pvolfiles` will be affected. But I guess the new alternative to `download_vpfiles`/`list_vp_files` will also be affected @PietrH ? I'm not sure how much lighter `httr` is but I guess to really cut dependencies we would  to not introduce that new import.

bart1:
Resolve #583 wait for lets wait for #579 and then resolve conflicts

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -26,7 +26,6 @@ Depends:
     R (>= 3.5.0)
 Imports:
     assertthat,
-    aws.s3,
     curl,
     fields,
     ggplot2,
@@ -52,6 +51,7 @@ Remotes:
     adokter/vol2birdR,
     bioc::rhdf5
 Suggests:
+    aws.s3,
     ggmap (>= 3.0.0),
     knitr,
     rlang,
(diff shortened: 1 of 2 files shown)


bart1:
As #579 is prepared for this ( https://github.com/adokter/bioRad/pull/579/commits/b2d42c3ec033fff967fa582b268eb30bd054d079 ). This can now be merged

adokter (approved):

PietrH (commented):
Agree to make `aws.s3` suggested, I'll make some changes to #579 to reduce merge conflicts.

DESCRIPTION
```diff
@@ -52,6 +51,7 @@ Remotes:
     adokter/vol2birdR,
     bioc::rhdf5
 Suggests:
+    aws.s3,
     ggmap (>= 3.0.0),
     knitr,
     rlang,
```
PietrH, line 57: Should rlang not be under `Imports` if we are using it to detect if `aws.s3` is installed?

Issue #148, opened 2018-09-25
Prepare package for rOpenSci onboarding


peterdesmet:
Guide: https://ropensci.github.io/dev_guide/index.html. Things I noticed that might not be covered yet:

- [ ] Add `codemeta.json` using `codemetar::write_codemeta()`
- [ ] Use `message()` and `warning()` to communicate with the user
- [ ] Code style, run `styler::style_pkg(filetype = c("R", "Rmd"))`
- [ ] README
- [ ] Add links for authors
- [ ] All packages should have a test suite that covers major functionality of the package. The tests should also cover the behavior of the package in case of errors.
- [ ] For http requests we strongly recommend using httr over RCurl.
- [ ] Try `goodpractice::gp()`

adokter:
You might want to check whether it is not a problem that our package uses an external Docker image that contains an important part of the code and functionality

peterdesmet:
Below is the full template for submitting an rOpenSci package for review. @adokter I suggest we edit this comment in preparation for submission. We should tackle all `TODO`s

---



Submitting Author Name: Adriaan Dokter
Submitting Author Github Handle: @adokter
Other Package Authors Github handles: (comma separated, delete if none) @peterdesmet, @github_handle2, TODO
Repository:  https://github.com/adokter/bioRad/
Version submitted: TODO
Submission type: Standard
Editor:  TBD 
Reviewers:  TBD 

Archive: TBD
Version accepted: TBD
Language: en

---



-   Paste the full DESCRIPTION file inside a code block below:

```
TODO
```


## Scope

- Please indicate which category or categories from our [package fit policies](https://ropensci.github.io/dev_guide/policies.html#package-categories) this package falls under: (Please check an appropriate box below. If you are unsure, we suggest you make a pre-submission inquiry.):

	- [x] data retrieval
	- [x] data extraction
	- [x] data munging
	- [ ] data deposition
    - [ ] data validation and testing
	- [ ] workflow automation
	- [ ] version control
	- [ ] citation management and bibliometrics
	- [ ] scientific software wrappers
	- [x] field and lab reproducibility tools
	- [ ] database software bindings
	- [x] geospatial data
	- [ ] text analysis

- Explain how and why the package falls under these categories (briefly, 1-2 sentences):

TODO

-   Who is the target audience and what are scientific applications of this package?

TODO

[...]

Issue #97, opened 2018-04-18
usage of `assertthat::assert_that` for input argument checks


stijnvanhoey:
Currently, Biorad uses `if (...) {stop(...)}` (with custom error message) and `stopifnot(...)` (without custom message) constructions to do input function validation. 

Making one-liners of the `if (...) {stop(...)}` construction is resulting in [long lines](https://github.com/adokter/bioRad/blob/rename_functions/R/nexrad_odim.R#L22) and making it all spread on multiple lines (cfr. syntax formatting) makes the functions _long_. 

Could we rewrite all of the input checks by using the `assertthat::assert_that(condition, msg =  "...")` functionality? This would give all input checks a similar result and get rid of all these short `if`-constructions.

adokter:
Yes I learnt of the assertthat package recently, and wished I had used that from the start. +1 for changing that (now, or in the future)

stijnvanhoey:
Check! I'm not taking it into account in the current renaming work, but let's put this on the enahncement list!

stijnvanhoey:
To have a safe roadmap and at the same time gain test-coverage: ;-)

1. create tests for the current input checks, make them work
2. rewrite with `assertthat`

adokter:
Issue #148 requires unit testing, therefore rewrite with `assertthat` likely part of rOpenSci onboarding

PietrH:
I had a quick look at how much work this would be, i think the `stopifnot()` assertations can pretty easily be replaced, we'll probably need to be a bit more careful with `stop()`. Writing tests first seems like the sensible thing to do. 

```r
#checking where we could replace assertations with assertthat::assert_that()
library(dplyr)

stopifnot_mentions <- system('grep "stopifnot(" -r .', intern = TRUE)
stop_mentions <- system('grep "stop(" -r . --include \\*.R', intern = TRUE)

count_mentions <- function(mentions) {
  
  readr::read_delim(
    paste(mentions, collapse = "\n"),
    col_names = c("file","mention")) %>%
    count(file,sort = TRUE) %>%
    print(n = Inf)
}

# should be easy replacements
count_mentions(stopifnot_mentions)
# need more review
count_mentions(stop_mentions)
```

For `stopifnot()`:
|file                                                              |  n|
|:-----------------------------------------------------------------|--:|
|./R/integrate_profile.R                                           | 10|
|./R/doy_noy.R                                                     |  8|
|./R/rcs.R                                                         |  6|
[...]

PR #588, opened 2023-05-26
Improve error test coverage

PietrH:
# STATUS:READY FOR REVIEW

In #97 it is proposed to replace current error generating code with `assertthat::assert_that()`. However, at the moment many error messages remain untested. This PR has the goal of improving the test coverage specifically for error messaging as a first step in an effort to #97. But is also related to #148 and the general effort to improve test coverage.

This PR doesn't fully resolve this observation, I believe there is still a lot of room for improvement when it comes to testing error messages. However, this PR can be merged and will offer a jumping off block for future efforts in improving our test coverage, as well as allowing work to start on #97.

I've tried to make improvements to error messages where I saw the opportunity. Some remain default messages, or descriptive. I suggest textual changes to the error messages themselves are tackled in a separate PR as to avoid discussions about how to phrase errors blocking this PR and #97 from going forward. 

Some functions remain entirely without tests. 

This PR improves test coverage, but due to it's focus on error messaging, little has changed when it comes to testing output objects and failure states other than errors. This is in scope of #148. 

### Todo:
- [x] ./integrate_to_ppi.R
- [x] ./scan_convert.R
- [x] ./bind_into_vpts.R
- [x] ./read_pvolfile.R
- [x] ./beam.R
- [x] ./composite_ppi.R
- [x] ./plot.vpts.R
- [x] ./read_vpts.R
- [x] ./map.R
- [x] ./operators.R
- [x] ./apply_mistnet.R
- [x] ./calculate_vp.R
- [x] ./download_vpfiles.R
- [x] ./integrate_profile.R
- [x] ./read_cajun.R
- [x] ./select.R
- [x] ./as.data.frame.R
- [x] ./download_basemap.R
- [x] ./plot.ppi.R
- [x] ./plot.scan.R
- [x] ./plot.vp.R
- [x] ./plot.vpi.R
- [x] ./rcs.R
- [x] ./regularize_vpts.R
- [x] ./sd_vvp_threshold.R
- [x] ./check_night.R
- [x] ./color_scale.R
- [x] ./get_param.R
- [x] ./get_quantity.R
- [x] ./read_vpfiles.R
- [x] ./doy_noy.R
- [x] ./vpts.R
- [x] ./get_elevation_angles.R
- [x] ./ppi.R
- [x] ./project_as_ppi.R
- [x] ./pvol.R
- [x] ./scan.R
- [x] ./vp.R
- [x] ./param.R
- [x] ./vpi.R

#### Todo bonus: 
- [ ] check if all assert_that() calls are covered by a test
- [x] check for `expect_error()` without `regex` arguments

# debt: issues I came across but couldn't or didn't fix

- 1x `stopifnot()` in ./scan_convert.R
- `read_cajun()` has impressive coverage but doesn't actually check output object all that well
- calculate_vp() mistnet install status is difficult to test
- `read_pvolfile()`: error messages on failing to read the HDF5 file, or errors within the file, missing data within the file... I struggled with creating a HDF5 file for testing.
- `read_vpfile()`: error `HDF5 file does not contain a /dataset1 group` can never trigger. Warning is possibly redundant.
- `read_vpts()` doesn't test a file that has a separator for the data
- `apply_mistnet()` doesn't test errors for local write permissions or mistnet installation status
- `rcs()` has a test that fails to ` assert_that: length of assertion is not 1`: however vectors seem to be allowed. I don't know what failure is expected, so I can't fix the error message.
- `integrate_to_ppi()` has error messages that remain untested, as well as error conditions that don't have custom errors (eg. passing a vector to `res`)
## testthat 3e

For this PR I upgraded to testthat 3rd edition, because I wanted to capture a more complex print message built by cat. This is quite a challenge in 2e but easy in 3e using a snapshot.

# General Notes
I've made some more specific notes as comments on this PR.

## Broken tests

I believe some of the error conditions in the code can't be naturally reached by users, I have made comments about these lines in this PR. 

I've found broken tests in `test-download_pvolfiles.R`, this wasn't caught because they failed to warnings instead of true failures. Mainly because the argument `fixed = true` was passed to `expect_error` even when no `regexp` was provided, or even when no error was expected `regexp = NA`

I've made an issue: #593, ~~I'm not sure if fixing all cases would be within the scope of this PR. To be discussed, we might want to split this effort off.~~ I've fixed this everywhere I could find it. I might have missed some.


tests/testthat/test-project_as_ppi.R
```diff
@@ -1,5 +1,15 @@
...
+  expect_error(
+    project_as_ppi.param(example_scan),
+    regexp = 'inherits(x, "param") is not TRUE',
+    fixed = TRUE
+  )
+  expect_error(
+    project_as_ppi.scan(get_param(example_scan, "DBZH")),
+    regexp = 'inherits(x, "scan") is not TRUE',
+    fixed = TRUE
+  )
```
PietrH, line 12: While these error states were uncovered, I do not believe a user could actually trigger these errors naturally. Either they pass a `scan` or a `param` object, or something else entirely. If they pass a `scan` or `param` object, no error should be returned, if they pass anything else, they should get; 

```
Error in UseMethod("project_as_ppi", x) : 
  no applicable method for 'project_as_ppi' applied to an object of class "{class_of_invalid_input_object}"
```

tests/testthat/test-bind_into_vpts.R
```diff
@@ -1,7 +1,38 @@
...
+  )
+  # non vps inputs
+  expect_error(
+    bind_into_vpts.vp(
+      example_vpts[2:5],
+      example_vpts[1]
+    ),
+    regexp = "requires vpts objects as input",
+    fixed = TRUE
+  )
```
PietrH, line 21: For a user to encounter this message, they would have to call bind_into_vpts.vp() directly, which I don't think is expected behaviour. 

As I understand it, a user either passes a supported object, or gets a different error entirely, for example in the case of a character vector: 
```

Error in UseMethod("bind_into_vpts", x) : 
  no applicable method for 'bind_into_vpts' applied to an object of class "character"
  ```

File: tests/testthat/test-download_basemap.R [hunk 3]
```diff
@@ -1,3 +1,16 @@
...
+    fixed = TRUE
+  )
+  # CRS absent in
+  ppi_no_crs <- ppi
+  raster::crs(ppi_no_crs) <- NA
+  expect_error(
+    download_basemap(ppi_no_crs),
+    regexp = "Not a projected ppi, download_basemap() expects a ppi generated by project_as_ppi() with argument project=TRUE",
+    fixed = TRUE
+  )
```
PietrH, line 15: I would have expected `download_basemap(project_as_ppi(example_scan, project = FALSE))` to also return this error, but it doesn't!

PietrH: fixed a typo in my test, so while github will flag this as outdated but I still suspect there might be a bug here in `download_basemap()`

tests/testthat/test-map.R
```diff
@@ -1,3 +1,52 @@
...
+  basemap_no_ppi <- basemap
+  attr(basemap_no_ppi,"ppi") <- FALSE
+  expect_error(
+    map(ppi, map = basemap_no_ppi),
+    regexp = "Not a ppi map, use download_basemap() to download a map.",
+    fixed = TRUE
+  )
+  # return error for basemap for different radar location
+  basemap_somewhere_else <- basemap
+  attributes(basemap_somewhere_else)$geo$lat <- 57
```
PietrH, line 46: It would be more defensive to also check for `lon`, but `map()` only checks `lat` test would fail.

tests/testthat/test-beam.R
```diff
@@ -122,3 +165,341 @@ test_that("Distance equals range when beam is following the earth (elev = 0)", {
...
+    fixed = TRUE
+  )
+  expect_error(
+    beam_width(range,
+               beam_angle = "a")
+  )
+})
+
+
+test_that("gaussian_beam_profile returns error on incorrect parameters", {
```
PietrH, line 373: Is this function actually used anywhere? It seems to be a wrapper for `gaussian_beam_profile_internal()` with better failstates. Is it possible this function was supposed to replace` gaussian_beam_profile_internal()` in other functions?

PietrH:
I'm pretty sure the error on [line 17](https://github.com/adokter/bioRad/blob/fb6b1886cb6a38cb75de4b959011f703287576b6/R/read_vpfiles.R#L17) of read_vpfiles() can never trigger: ` stop("HDF5 file does not contain a /dataset1 group")`

Because within the warning state above: 

```R

  if (!is.vpfile(file)) {
    warning(paste(file, "is not a vertical profile"))
    return(NULL)
  }

```
`is.vpfile()` will [call](https://github.com/adokter/bioRad/blob/fb6b1886cb6a38cb75de4b959011f703287576b6/R/vpfile.R#L25) `get_odim_object_type()` [which in turn](https://github.com/adokter/bioRad/blob/fb6b1886cb6a38cb75de4b959011f703287576b6/R/pvolfile.R#L60) has error handling by `is.odimfile()`:

```R
get_odim_object_type <- function(file) {
  if (!file.exists(file)) {
    warning(glue::glue("Can't find {file}"))
    return(NA)
  }
  if (!is.odimfile(file)) {
    # Errors are handled by is.odimfile()
    return(NA)
  }
  ```
  
  And `is.odimfile()` generates warnings for this state, but will also cause a `FALSE` in `is.vpfile()` meaning that the underlying lines are never read, and the error state within `read_vpfile()` will never trigger:
  
  ```R
  is.odimfile <- function(file) {
  if (!rhdf5::H5Fis_hdf5(file)) {
    warning(glue::glue("{file} is not an HDF5 file."))
    return(FALSE)
  }
  output <- TRUE
  groups <- rhdf5::h5ls(file, recursive = FALSE)$name
  if (!("dataset1" %in% groups)) {
    output <- FALSE
    warning(
      glue::glue("HDF5 file {file} does not contain a `/dataset1` group.")
    )
  }

  if (!("what" %in% groups)) {
    output <- FALSE
    warning(
      glue::glue("HDF5 file {file} does not contain a `/what` group.")
    )
  } else {
    object <- rhdf5::h5readAttributes(file, "what")$object
    if (is.null(object)) {
      warning(
        glue::glue("HDF5 file {file} does not contain an `object` attribute in the ",
        "`/what` group.")
      )
      output <- FALSE
    }
  }

  if (!("how" %in% groups)) {
    # accepting a missing /how group
    warning(
      glue::glue("HDF5 file {file} does not contain a `/how` group.")
    )
  }
  if (!("where" %in% groups)) {
    output <- FALSE
    warning(
      glue::glue("HDF5 file {file} does not contain a `/where` group.")
    )
  }
  return(output)
}
```

tests/testthat/test-rcs.R
```diff
@@ -5,17 +5,31 @@ vpts <- example_vpts
...
-  expect_error(rcs(vp) <- "not_numeric")
-  expect_error(rcs(vp) <- NULL)
+  expect_error(rcs(vp) <- "not_numeric",
+               regexp = "value is not a numeric or integer vector",
+               fixed = TRUE)
+  expect_error(rcs(vp) <- NULL,
+               regexp = "value is not a numeric or integer vector",
+               fixed = TRUE)
   expect_error(rcs(vp) <- c(2, 2))
   expect_error(rcs("not_a_vp") <- 5)
```
PietrH, line 26: These two expectations are succeeding, because they result in tests. However, the messages are confusing. 

I'm unsure of what the expectations actually are here, so I couldn't fix them. 

```R
> rcs(vp) <- c(2, 2)
#> Error: assert_that: length of assertion is not 1
> rcs("not_a_vp") <- 5
#> Error in rcs("not_a_vp") <- 5 : 
#>  target of assignment expands to non-language object
```

From the lines above it seems to me integer vectors should be allowed. The second expectation should probably be `no applicable method for 'rcs' applied to an object of ...` or an improvement upon the default.

tests/testthat/test-sd_vvp_threshold.R
```diff
@@ -4,16 +4,37 @@ vp_list_mixed <- list(example_vp, "not_a_vp")
...
+    sd_vvp_threshold(vp) <- NULL,
+    regexp = "value is not a numeric or integer vector",
+    fixed = TRUE
+    )
+  expect_error(
+    sd_vvp_threshold(vp) <- c(2, 2)
+    )
+  expect_error(
+    sd_vvp_threshold("not_a_vp") <- 2
+    )
```
PietrH, line 32: Similar issue as in `test-rcs()` see https://github.com/adokter/bioRad/pull/588#discussion_r1217815890

R/integrate_to_ppi.R
```diff
@@ -157,18 +157,54 @@ integrate_to_ppi <- function(pvol, vp, nx = 100, ny = 100, xlim, ylim, zlim = c(
...
+      is.numeric(xlim),
+      msg = "'xlim' should be an integer vector of length two")
+    assertthat::assert_that(
+      length(xlim) == 2,
+      msg = "'xlim' should be an integer vector of length two")
+    assertthat::assert_that(
+      all(!is.na(xlim)),
+      msg = "'xlim' should be a vector with two numeric values for upper and lower bound")
+    assertthat::assert_that(
+      xlim[1] < xlim [2],
```
PietrH, line 172: These all seemed flipped compared to the examples, so I switched them over from `xlim[1] > xlim[2]` to `xlim[1] < xlim[2]`

Please review carefully. The same for `xlim` and `zlim`. What should happen if they are the same (range is 0)?

R/integrate_to_ppi.R
```diff
@@ -389,13 +436,6 @@ add_expected_eta_to_scan <- function(scan, vp, quantity = "dens",
     stop(paste("radar longitude cannot be found in polar volume,",
                "specify using 'lon' argument"))
   }
-  if (is.null(scan$geo$height) &&
-      missing(antenna)) {
-    stop(
-      paste("antenna height cannot be found in polar volume,",
-            "specify antenna height using 'antenna' argument")
-    )
-  }
```
PietrH, line 398: Repeat from L368-374 (old) or L415-421 (new)

R/integrate_profile.R
```diff
@@ -189,9 +189,11 @@ integrate_profile <- function(x, alt_min, alt_max,
...
                                  interval_max = 3600, interval_replace = NA, height_quantile = NA) {
   stopifnot(inherits(x, "vp"))
-  stopifnot(is.numeric(alt_min) | alt_min=="antenna")
-  stopifnot(is.numeric(alt_max))
-  stopifnot(is.na(alpha) || is.numeric(alpha))
+  assertthat::assert_that(
+    is.numeric(alt_min) || alt_min=="antenna"
+  )
+  assertthat::assert_that(assertthat::is.number(alt_max))
+  assertthat::assert_that(is.na(alpha) || is.numeric(alpha))
```
adokter, line 196: `is.numeric` should be `is.number`, alpha cannot be a vector

R/plot.vpts.R
```diff
@@ -105,7 +105,10 @@ plot.vpts <- function(x, xlab = "time", ylab = "height [m]", quantity = "dens",
                       n_color=1000, palette = NA,
                       ...) {
   stopifnot(inherits(x, "vpts"))
-  stopifnot(quantity %in% names(x$data))
+  assertthat::assert_that(
+    quantity %in% names(x$data),
+    msg = "quantity needs to be one of `dens`, `eta`, `dbz` or `DBZH`"
```
adokter, line 110: this message is too specific, the user can plot any of the available quantities, e.g. also speed, direction etc. Message should simply warn if user request a quantity that does not exist.

R/read_pvolfile.R
```diff
@@ -99,23 +99,27 @@ read_pvolfile_body <- function(file, param = c(
...
-    stop("'sort' should be logical")
-  }
+  assertthat::assert_that(assertthat::is.flag(sort),
+                          msg = "'sort' should be logical")
   if (!missing(lat)) {
-    if (!is.numeric(lat) || lat < -90 || lat > 90) {
-      stop("'lat' should be numeric between -90 and 90 degrees")
-    }
+    lat_msg <- "'lat' should be numeric between -90 and 90 degrees"
+    assertthat::assert_that(length(lat) == 1, msg = lat_msg)
```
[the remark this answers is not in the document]
adokter: same for new line 113 and 120

R/read_vpts.R
```diff
@@ -30,17 +30,33 @@ read_vpts <- function(file, radar, lat, lon, height, wavelength = "C", sep="") {
...
+                          msg = "'sep' should be either \",\" or \"\"")
+  assertthat::assert_that(sep == "" || sep == ",",
+                          msg = "'sep' should be either \",\" or \"\"")
+
   if (!missing(lat)) {
+    assertthat::assert_that(
+      assertthat::is.number(lat),
+      msg = "'lat' should be a single numeric between -90 and 90 degrees"
+      )
     if (!is.numeric(lat) || lat < -90 || lat > 90) {
```
adokter, line 45: `!is.numeric(lat)` can be dropped here

R/read_vpts.R
```diff
@@ -30,17 +30,33 @@ read_vpts <- function(file, radar, lat, lon, height, wavelength = "C", sep="") {
...
       stop("'lat' should be numeric between -90 and 90 degrees")
     }
   }
   if (!missing(lon)) {
-    if (!is.numeric(lon) || lat < -360 || lat > 360) {
+    assertthat::assert_that(
+      assertthat::is.number(lon),
+      msg = "'lon' should be a single numeric numeric between -360 and 360 degrees"
+    )
+    if (!is.numeric(lon) || lon < -360 || lon > 360) {
```
adokter, line 54: `!is.numeric(lon)` can be dropped here

R/read_vpts.R
```diff
@@ -54,19 +70,24 @@ read_vpts <- function(file, radar, lat, lon, height, wavelength = "C", sep="") {
...
   }
+  wavelength_msg <-
+    glue::glue(
+      "'wavelength' should be a single positive number",
+      ", or one of 'C' or 'S' for C-band and S-band radar, respectively."
+    )
+  assertthat::assert_that(length(wavelength) == 1,
+                          msg = wavelength_msg)
+  assertthat::assert_that(
+    (is.numeric(wavelength) && wavelength > 0) ||
```
adokter, line 81: if you change `is.numeric(wavelength)` to `is.number(wavelength)` line 78/79 can be dropped

tests/testthat/test-plot.vpts.R
```diff
@@ -32,6 +41,70 @@ test_that("plot.vpts() warns for deprecated arguments", {
...
+    suppressMessages(plot(regularize_vpts(example_vpts), palette = 123)),
+    regexp = "palette should be a character vector with hex color values",
+    fixed = TRUE
+  )
+})
+
+test_that("plot_wind_barbs() returns error on incorrect parameters", {
+  expect_error(
+    plot_wind_barbs(c(1, 2), c(4, 5, 8)),
+    regexp = "X AND Y COORDINATES SHOULD HAVE SAME LENGTH!",
```
adokter, line 55: we should change to lower case here and in the function

tests/testthat/test-plot.vpts.R
```diff
@@ -32,6 +41,70 @@ test_that("plot.vpts() warns for deprecated arguments", {
...
+    fixed = TRUE
+  )
+
+  variable_error_message <-
+    "ALL VARIABLES SHOULD HAVE SAME LENGTH AS COORDINATES, OR BE MISSING!!!"
```
adokter, line 60: change to lower case and fewer exclamation marks

tests/testthat/test-project_as_ppi.R
adokter: true, should test only against `project_as_ppi()` and not its class methods. We could strip away the .param and .scan postfix to the main function

tests/testthat/test-read_vpts.R
```diff
@@ -1,3 +1,180 @@
...
+    regexp = "No 'wavelength' argument provided, assuming radar operates at C-band",
+    fixed = TRUE
+  )
+  # test a vptsfile with a comma seperator
+  vptsfile_lines <- readLines(vptsfile)
+  vptsfile_comment_lines <- grepl("#", readLines(vptsfile))
+  vptsfile_lines[!vptsfile_comment_lines] <-
+    gsub("\\s+", ",", readLines(vptsfile)[!vptsfile_comment_lines])
+  vptsfile_sep <- tempfile()
+  writeLines(vptsfile_lines, vptsfile_sep)
```
adokter, line 179: incomplete?

adokter (changes requested):

R/integrate_profile.R
PietrH: If I make this change, the following tests fail: 
```R
expect_s3_class(integrate_profile(c(example_vp,example_vp)), "vpi")
```

https://github.com/adokter/bioRad/blob/c87ec9d5caa23a4789c56408a9aa2537193959dd/tests/testthat/test-integrate_profile.R#L3

And 

```R
expect_error(
    integrate_profile(example_vp, alt_min = 500, alt_max = 300),
    regexp = "'alt_min' should be smaller than 'alt_max'",
    fixed = TRUE
    )
```

Can you see what is going wrong here?

tests/testthat/test-read_vpts.R
PietrH: Well spotted, a train of thought I didn't complete. Testing a file with a different separator is something to consider for the future.

R/integrate_profile.R
PietrH: Found it, need to allow NA

PietrH:
I believe I've addressed all requested changes. Could you have another look at this @adokter ?

adokter (approved):

Issue #551, opened 2023-03-10
Adapt `read_vpts()`and VPTS object to VPTS CSV format


peterdesmet:
## [Current functionality](https://adriaandokter.com/bioRad/reference/read_vpts.html)

Read a time series of vertical profiles (`vpts`) from file

```R
read_vpts(
  file,
  radar,
  lat,
  lon,
  height,
  wavelength = "C"
)
```

## Reason for change

- [VPTS CSV](https://aloftdata.eu/vpts-csv/) is now a standardized format for VPTS data. We need a function to read that data
- It is intended to replace the "A text file containing the standard output (stdout) generated by vol2bird (or the package function `calculate_vp`)."

## Questions

- [ ] Should data be converted to VPTS object or remain a data frame? This depends on whether we want to keep vpts objects around or make those regular data frames. @adokter 
- [ ] VPTS objects have some conventions regarding column order and names. I think those should be adopted in VPTS (data frames).
- [ ] Should it be possible to read from a URL in addition to single file path (supported by readr)
- [ ] Should we still support the legacy `stdout` objects? That requires keeping around all the parameters too.

## Todo

- [ ] Remove `lat` parameter. It is included in data as `radar_latitude`
- [ ] Remove `long` parameter. It is included in data as `radar_longitude`
- [ ] Remove `height` parameter. It is included in data as `radar_height`
- [ ] Remove `wavelength` parameter. It is included in data as `radar_wavelength`
- [ ] Allow to read a list of file paths, rather than single `file`. Those files will be combined.
- [ ] Add `readr` as dependency

Issue #555, opened 2023-04-05
Create `as.vpts()` function


peterdesmet:
It is currently possible to convert a `vpts` object to a dataframe with `as.data.frame()`. The reverse is not possible, i.e. when you have a dataframe with vpts data, you cannot create a `vpts` object (to plot or use for other functions).

In the short term, it would be good to have such a function `create_vpts(df)`. This function could then also be used by the new `read_vpts()` function (#551).

In the longer term, `vpts` objects should become data frames (rather than currently lists), so it is easier to use e.g. `dplyr::filter()` and import data from vpts csv without having to convert them to objects.

bart1:
Quickly tagging @jedgroev as he might be intrested here

adokter:
included in #590

PR #590, opened 2023-05-29
read_vpts()

peterdesmet:
This PR is incomplete, but contains the start of a new `read_vpts()` function that supports both hdf5 vp files and VPTS CSV files (#551). It would be very useful if the latter was included in bioRad 0.7.

- The public `read_vpts(files)` function checks if all provided file extensions are the same and then uses a `switch()` to read the data with an appropriate helper function: either `read_vpts_csv()` or `read_vpts_hdf5()`.
  - `read_vpts_csv()` contains new functionality, but relies heavily on [frictionless](https://cran.r-project.org/package=frictionless) to read local and remote csv, gzipped or not (all supported by underlying `readr` dependency) and compares the data with the VPTS CSV table schema. It provides good warning/errors.
  - `read_vpts_hdf5()` makes use of `read_vpfiles()` and then `bind_into_vpts()`.
- A previous function with the name `read_vpts()` has been renamed to `read_stdout()`. It is not integrated in the new `read_vpts()` because 1) that  read function requires a lot more metadata parameters (`lat`, `lon`, `height`) that I did not want to integrate in the new `read_vpts()` and 2) support for vol2bird stdout format will be deprecated in the future.
- A helper script `/tests/testthat/helper.R` has been added to aid in the download and temporary storage of hdf5 and csv files required for testing.   When tests are run, a temporary folder `/temp` is created and subsequently removed at the end of the `test-read_vpts.R` script.   

## TODO

- [x] Provide documentation
- [x] Create a public `read_vpts()` function
- [x] Create a private `read_vpts_hdf5()` function
- [x] Create a private `read_vpts_csv()` function
- [x] `read_vpts_csv()` can return data as df (useful for the future #568)
- [x] `read_vpts_csv()` can return data as a vpts object. A start for this has been made, code can take inspiration from `read_stdout()`
- [x] Write tests for `read_vpts()`. Necessary tests are listed, but not written
- [x] Write tests for `read_stdout()`. These were not provided yet, but are not required as part of this PR
- [x] Add deprecation rerouting to `read_stdout()` if `read_vpts()` is used with parameters `lat`, `lon`, etc.

Unfortunately, I don't have the time to finish this. @iskandari would have the opportunity to complete this functionality? Ping me if you have any questions.


iskandari:
Absolutely!

tests/testthat/test-read_vpts.R
```diff
@@ -1,3 +1,85 @@
...
+    file.remove(temp_file)
+  }
+})
+
+
+test_that("read_vpts() returns error on multiple radars in vp hdf5 files", {
+
+})
+
+test_that("read_vpts() can read local (gzipped) VPTS CSV files", {
```
iskandari, line 67: @peterdesmet for reading local files, shouldn't they first be brought in with curl, stored temporarily, and then removed?  In essence, it's just reading from a remote URL and then storing locally, correct?

peterdesmet: Yes, as part of the test preparation (first lines of `test-read_vpts.R`) some remote files will need to be downloaded, stored in a temp file and their path saved in a variable. They can then be used for this test (read local files) and then removed from the temp dir.

Thinking about it, it might be good to make a test-helper-function to download files and to cleanup files, e.g. in a `tests/testthat/helper-download.R`. See this blog post: https://blog.r-hub.io/2020/11/18/testthat-utility-belt/

iskandari:
@peterdesmet have a look at the changes added and lmk if anything remains to be done; otherwise, this is ready for review

tests/testthat/helper.R
```diff
@@ -0,0 +1,43 @@
+# Define the URLs
+urls <- c(
+  "https://lw-enram.s3-eu-west-1.amazonaws.com/be/jab/2016/09/19/23/bejab_vp_201609192315.h5",
```
peterdesmet, line 3: Don't use lw-enram bucket, it will be deleted. The equivalent file is:

```
https://lw-enram.s3-eu-west-1.amazonaws.com/ecog-04003/hdf5/bejab/2016/09/19/bejab_vp_201609192315.h5
```

But it might be better to use an equivalent of the vpts file from 2023

tests/testthat/helper.R
```diff
@@ -0,0 +1,43 @@
...
+}
+
+# Create the h5 and csv sub-directories
+h5_dir <- file.path(temp_dir, "h5")
+csv_dir <- file.path(temp_dir, "csv")
+if (!dir.exists(h5_dir)) dir.create(h5_dir)
+if (!dir.exists(csv_dir)) dir.create(csv_dir)
+
+# Function to download a file
+download_test_file <- function(url, dest_dir = temp_dir) {
```
peterdesmet, line 27: This helper.R file should only contain the `download_test_file()` function. Any code above and below is in preparation of the `read_vpts()` tests and should be moved to `test-read_vpts.R`, where the helper function can be called.

R/read_vpts.R
```diff
@@ -1,222 +1,124 @@
...
   output <- list(
-    radar = radar, datetime = datetime, height = heights,
-    daterange = .POSIXct(c(min(datetime), max(datetime)), tz = "UTC"),
-    timesteps = difftimes, data = vpsFlat,
-    attributes = attributes, regular = regular
+    radar = radar,
+    datetime = datetime,
+    height = heights,
+    daterange = c(min(datetime), max(datetime)),
+    # timesteps = difftimes,
```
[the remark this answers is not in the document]
adokter: @peterdesmet this suggest one of the elements in your `data$timesteps` is NA, which shouldn't happen - which file did you load here?

tests/testthat/helper.R
iskandari: Note: choose something from aloft - aloft.s3-eu-west-1.amazonaws.com

R/read_vpts.R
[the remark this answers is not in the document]
peterdesmet:
``` r
library(bioRad)
#> Welcome to bioRad version 0.7.0.9590
#> using vol2birdR version 0.2.2.9043 (MistNet not installed)
#> Assigning sp_evolution_status to 2. See sp::get_evolution_status()
#> This is required until the 'sp' package deprecates 'rgdal'
files <- c(
  "https://aloft.s3-eu-west-1.amazonaws.com/baltrad/monthly/bejab/2023/bejab_vpts_202303.csv.gz",
  "https://aloft.s3-eu-west-1.amazonaws.com/baltrad/monthly/bejab/2023/bejab_vpts_202304.csv.gz"
)
vpts <- read_vpts(files)
vpts
#>                    Irregular time series of vertical profiles (class vpts)
#> 
#>            radar:  1 
#>       # profiles:  5029 
#> time range (UTC):  2023-03-03 22:00:00 - 2023-04-30 23:45:00 
#>    time step (s):  min: NA     max:  NA
```

<sup>Created on 2023-06-08 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

tests/testthat/helper.R
peterdesmet: Indeed, I wrote the wrong link, now corrected.

R/read_vpts.R
```diff
@@ -1,222 +1,124 @@
...
-    wavelength <- 5.3
-  }
-  if (wavelength == "S") {
-    wavelength <- 10.6
-  }
-  if (!is.numeric(wavelength) || length(wavelength) > 1) {
-    stop("Not a valid 'wavelength' argument.")
+read_vpts <- function(files, radar = NULL, lat = NULL, lon = NULL, height = NULL) {
+  # Check if any of the old parameters are used
+  if (!is.null(radar) | !is.null(lat) | !is.null(lon) | !is.null(height)) {
```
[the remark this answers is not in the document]
peterdesmet: If we want to provide backwards compatibility, then yes. I'm not sure how often people used `read_vpts()` to be honest.

R/read_vpts.R
```diff
@@ -1,241 +1,124 @@
...
-  tz = "UTC"
-  )
-  data <- lapply(
-    data,
-    function(x) {
-      x["datetime"] <- NULL
-      x["profile_index"] <- NULL
-      x
-    }
+  # The following steps convert the data to a vpts object
```
bart1, line 72: Would it be an idea to put this data frame to vpts conversion into a separate function, that would enable reuse by people that get the data for example out of a database. I'm not sure as this also touches on the new vpts format. This might also relate to #555

peterdesmet: Yeah, but it becomes moot the moment we move vpts objects to dataframes _in_ bioRad (#568).

R/read_vpts.R
```diff
@@ -1,222 +1,124 @@
...
-  }
 
-  # add profile_index to identify consecutive profiles
-  data$new_profile_starts <- c(T, (data$height[-1] - data$height[-length(data$height)]) < 0)
-  data$profile_index <- NA
-  profile_index <- NULL # define profile_index to suppress devtools::check warning in next line
-  data[which(data$new_profile_starts), "profile_index"] <- 1:length(which(data$new_profile_starts))
-  data <- tidyr::fill(data, profile_index)
+  # Read files
+  data <- switch(extension,
```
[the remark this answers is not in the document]
iskandari: Added a `guess_filet_ype()` helper function to interpret files without extensions.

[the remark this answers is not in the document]
peterdesmet: Nice! Some remarks:

1. Does it work on multiple files? The `file_path` argument seems to be singular, but multiple `files` are passed.
2. There doesn't seem to be a check if all provided extensions are the same when file type is guessed

Can I suggest:

1. To maybe name the function `get_extension()`
2. As first part of the function, try `tools::file_ext(files)`
3. If that fails, to guess. Is there a performance downside for doing so?
4. To not assume `txt`, but return `NA` if the extension cannot be guessed
5. To return a vector with the file extensions, e.g. `c("txt", "txt", "csv", "NA")`

Outside the function, in `read_vpts()`:

1. To get the unique extensions: `extension <- unique(get_extension(files))`
2. If we want so, to convert `NA` to `txt` (i.e. assume txt if no extension was found)
3. To still check only one extension was returned (= current `assertthat::assert_that(length(extension) == 1`)

I think that would be more defensive and DRY

R/read_vpts.R
```diff
@@ -1,241 +1,144 @@
...
+#' The following file formats are supported (but cannot be mixed):
+#' - [VPTS CSV](https://aloftdata.eu/vpts-csv/).
+#' - [ODIM bird profile](https://github.com/adokter/vol2bird/wiki/ODIM-bird-profile-format-specification).
+#' @param files Path(s) to one or more files containing vpts data.
+#' @param ... Additional arguments for backward compatibility, passed to `read_stdout`.
+#' @return `vpts` object.
+#' @family read functions
 #' @export
-#'
-#' @examples
```
adokter, line 19: each function needs examples demonstrating most of the functionality

R/read_vpts.R
```diff
@@ -1,241 +1,144 @@
...
-  if (wavelength == "C") {
-    wavelength <- 5.3
-  }
-  if (wavelength == "S") {
-    wavelength <- 10.6
+  # Check if the input file has a .txt extension and if so reroute to read_stdout
+  if (extension == "txt") {
+    warning(".txt extenstion detected - falling back to read_stdout(). The use of read_stdout() will be deprecated soon.
+    Please consider updating your code to use csv or h5 input files")
+    return(do.call(read_stdout, c(list(file = files, wavelength = "C", sep = ""), list(...))))
```
adokter, line 55: arguments are not parsed correctly, user should be able to specify the arguments available in `read_stdout()`, see this error:
```
stdout_file <- system.file("extdata", "example_vpts.txt", package = "bioRad")
read_vpts(stdout_file, radar = "KBGM", wavelength = "S")
```
produces:
```
Error in (function (file, radar, lat, lon, height, wavelength = "C", sep = "")  : 
  formal argument "wavelength" matched by multiple actual arguments
```

R/read_vpts.R
```diff
@@ -1,241 +1,144 @@
...
-    msg = wavelength_msg)
 
-  if (wavelength == "C") {
-    wavelength <- 5.3
-  }
-  if (wavelength == "S") {
-    wavelength <- 10.6
+  # Check if the input file has a .txt extension and if so reroute to read_stdout
+  if (extension == "txt") {
+    warning(".txt extenstion detected - falling back to read_stdout(). The use of read_stdout() will be deprecated soon.
```
adokter, line 53: I don't think we need to deprecate read_stdout, we can keep the functionality for quite a while. 

Suggested behaviour for .txt extension:
1. verify if comma delimited or not
2. if comma delimited, try reading VPTS CSV
3. if not comma delimited, try reading stdout format

So allow user to name their csv file with a txt extension or without any extension

R/read_vpts.R
```diff
@@ -1,241 +1,144 @@
...
-    "sd_vvp", "gap", "dbz", "eta", "dens", "DBZH", "n",
-    "n_dbz", "n_all", "n_dbz_all"
-  )
-  header.names.long <- c(
-    "Date", "Time", "height", "u", "v", "w", "ff", "dd",
-    "sd_vvp", "head_bl", "head_ff", "head_dd", "head_sd",
-    "gap", "dbz", "eta", "dens", "DBZH", "n", "n_dbz",
-    "n_all", "n_dbz_all"
+  # Read files
+  data <- switch(extension,
```
adokter, line 59: suggest renaming `extension` to `file_type` since we will allow text files without extension

R/read_vpts.R
adokter: @peterdesmet @iskandari Suggest adding a short VPTS CSV file in `/inst/extdata` that we can use in examples. Helpful if you pick an example file that also has some interesting bird profile in it, so it can be used in other examples as well. Can be a single night with a little lead and end time to keep the package small.

adokter: Added example prototypes in 6d2311d7edb02061f09c934ca9dbdb7e81df1874, these have several FIXMEs that still need to be addressed

R/read_vpts.R
```diff
@@ -1,241 +1,144 @@
...
-                          msg = "'sep' should be either \",\" or \"\"")
-
-  if (!missing(lat)) {
-    assertthat::assert_that(
-      assertthat::is.number(lat),
-      msg = "'lat' should be a single numeric between -90 and 90 degrees"
+read_vpts <- function(files, ...) {
+
+  #Define valid extensions
+  valid_extensions <- c("csv", "gz", "h5", "txt")
```
adokter, line 27: suggest allowing no extension

R/read_vpts.R
```diff
@@ -1,241 +1,144 @@
...
+#'
+#' @inheritParams read_vpts
+#' @param df If `TRUE` returns data as data frame rather than `vpts`
+#'   object.
+#' @return `vpts` object.
+#' @noRd
+read_vpts_csv <- function(files, df = FALSE) {
+  # Create Frictionless Data Package
+  package <- frictionless::create_package()
+  schema <- "https://raw.githubusercontent.com/enram/vpts-csv/main/vpts-csv-table-schema.json"
```
adokter, line 78: does this mean we can only read files when there is an internet connection? I think we want read_vpts to work also in absence of internet.

R/read_vpts.R
```diff
@@ -1,241 +1,144 @@
...
-#' [`vpts()`][summary.vpts] for details.
+#' Read time series of vertical profiles (`vpts`) from file(s)
 #'
+#' Reads `vpts` data from one or more files.
+#' The following file formats are supported (but cannot be mixed):
+#' - [VPTS CSV](https://aloftdata.eu/vpts-csv/).
+#' - [ODIM bird profile](https://github.com/adokter/vol2bird/wiki/ODIM-bird-profile-format-specification).
+#' @param files Path(s) to one or more files containing vpts data.
+#' @param ... Additional arguments for backward compatibility, passed to `read_stdout`.
+#' @return `vpts` object.
```
adokter, line 9: currently only reads of stdout and h5 formats returns a true `vpts`object, the vpts csv function returns a `data.frame` that is not a true bioRad `vpts` object. So we need to document in which cases the return value is a `vpts` and in which cases it is a `data.frame`

R/read_vpts.R
adokter: Just reading that in tests you download VPTS CSV files. You could do the same in the example (instead of including file in `/inst/extdata`, but then that part of the example should be in a \dontrun{} block, because we can't assume internet connectivity

adokter (changes requested):

R/read_vpts.R
peterdesmet: Correct. If we want to support no connection, we need to store that schema locally in bioRad.

R/read_vpts.R
peterdesmet: The goal (for now) is to always return a true bioRad `vpts` object.

R/read_vpts.R
peterdesmet: Indeed, you can read remotely, so no need to store locally. You can look on https://crow.aloftdata.eu/#/ for a profile of interest

iskandari:
@peterdesmet is it possible to validate values in a dataframe against the json table schema with `frictionless`? 

For example, I imported a VPTS CSV, changed values in `datetime` and`height` to violate constraints, and tried using the [check_schema()](https://github.com/frictionlessdata/frictionless-r/blob/main/R/check_schema.R) function to validate but this returned TRUE.   
```
library(frictionless)

system("curl https://aloft.s3-eu-west-1.amazonaws.com/baltrad/monthly/bejab/2023/bejab_vpts_202304.csv.gz > sample.csv.gz")
system("gunzip sample.csv.gz")
my_df = read.csv('sample.csv')

#two cols violate constraints
my_df$datetime = 1
my_df$height = 50000

package <- frictionless::create_package()
schema <- "https://raw.githubusercontent.com/enram/vpts-csv/main/vpts-csv-table-schema.json"
package <- frictionless::add_resource(
  package,
  "vpts",
  data = my_df,
  schema = schema
)

# Read resource (compares data with schema and binds rows of all files)
df <- frictionless::read_resource(package, "vpts")


#' Check Table Schema object
#'
#' Check if an object is a list describing a Table Schema and (optionally)
#' compare against a provided data frame.

check_schema(package$resources[[1]]$schema, my_df)
[1] TRUE
```

peterdesmet:
Correct, there is no full validation in frictionless. `check_schema()` will check if the provided schema is well-formed and aligns with the columns in the data. And if you would add integers in e.g. `datatime`, then `readr::read_csv()` that is used under the hood will change those to NA and warn about it (can be inspected with `problems()`).

For full validation (including constraints), you currently need Frictionless Framework, e.g. from the command line `frictionless validate datapackage.json`. I can provide more info if needed, but I don't think this functionality should be build into `read_vpts()`. As with the frictionless package, it's going to do a best attempt at reading the data and warn about `problems()`.

R/read_vpts.R
iskandari: `read_vpts()` should now always return a true `vpts` object

iskandari: `read_vpts()` should now always return a true `vpts` object even from csv

tests/testthat/test-calculate_vp.R
```diff
@@ -208,7 +208,7 @@ test_that("calculate_vp() produces a vp object and optional vpfile, pvolfile", {
   vp_from_file <- read_vpfiles(vpfile)
   pvol_from_file <- read_pvolfile(pvolfile_out)
 
-  expect_s3_class(vp, "vp")
+  expect_s3_class(vp, "vpts")
```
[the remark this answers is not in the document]
adokter: Noticing a related bug in how `calculate_vp()`, the function now returns a vpts object of lenght 1, while it should return a vp object.
```
pvolfile <- system.file("extdata", "volume.h5", package = "bioRad")
calculate_vp(pvolfile)
```

tests/testthat/test-read_vpts.R
```diff
@@ -1,173 +1,233 @@
...
+  "https://aloft.s3-eu-west-1.amazonaws.com/baltrad/monthly/bejab/2023/bejab_vpts_202303.csv.gz",
+  "https://aloft.s3-eu-west-1.amazonaws.com/baltrad/monthly/bejab/2023/bejab_vpts_202304.csv.gz",
+  "https://aloft.s3-eu-west-1.amazonaws.com/baltrad/monthly/bewid/2023/bewid_vpts_202303.csv.gz"
+)
 
-test_that("read_vpts() returns error on incorrect parameters", {
-  empty_file <- tempfile()
-  file.create(empty_file)
+# Define the path to the new temporary directory
+temp_dir <- "temp"
```
adokter, line 12: Best practice is to use R function `tempdir()` for defining temporary directories, not creating your own temp directory in the user space

tests/testthat/test-read_vpts.R
```diff
@@ -1,173 +1,233 @@
...
-  expect_warning(
-    read_vpts(vptsfile,
-      radar = "KBGM"
-    ),
-    regexp = "No 'wavelength' argument provided, assuming radar operates at C-band",
-    fixed = TRUE
+# Comapre read_vpts output from data in both formats
+
+if (!require("aws.s3")) {
+  stop("Package 'aws.s3' is not installed. Please install it before proceeding.")
```
adokter, line 171: aws.s3 is in Suggests in DESCRIPTION, so we can expect it to be present on test servers. Instead I would skip these tests when it is not available.

R/calculate_vp.R
```diff
@@ -388,7 +392,7 @@ calculate_vp <- function(file, vpfile = "", pvolfile_out = "",
   )
 
   # read output into a vp object
-  output <- read_vpfiles(profile.tmp)
+  output <- read_vpts(profile.tmp)
```
adokter, line 395: output should be a vp object, not vpts, even when the user has specified a .csv file as output.

So you should keep `read_vpfiles()`, and modify the output format from vpts in the case of a .csv file output. You can do that quite easily by subsetting the output format: when profile.tmp is a csv file, you will get a vpts object of lenght 1 from `read_vpfiles()`, which you can convert to vp using `output[1]`. bioRad has the [] operator overloaded, such that it can subset vpts objects.

DESCRIPTION
```diff
@@ -30,27 +30,29 @@ Imports:
...
     lubridate,
     lutz,
-    maptools,
+    suntools,
     methods,
     raster,
     readr,
     dplyr (>= 1.1.0),
     rhdf5,
     sp,
```
adokter, line 45: FYI latest sp version 2 now has sp_evolution_status to 2 by default, but version two is only available for the latest R version. I think requiring sp >= 2.0 is too strict, as that will force too many folks to update their base R

adokter (changes requested):

tests/testthat/test-calculate_vp.R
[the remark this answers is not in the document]
iskandari: Note: `vp` and `vpts` objects are each lists containing a `data` object.   The main difference is that `vp` data objects are of class `data.frame` whereas `vpts` objects are a list of lists.  Seeing as `vp` objects represent data at a single nominal timestamp, beam height is included in the data frame itself as opposed to a separate vector.

R/calculate_vp.R
iskandari: There is also the helper function `vpts_to_vp()` which allows us to keep `read_vpts()` as is and then convert back to `vp`

iskandari (approved):

R/calculate_vp.R
iskandari: Addressed in e97ef0768f19906f14781edbaf41dd1bbb83e2b9

tests/testthat/test-read_vpts.R
iskandari: FIxed in 62eaac0a0183db77771c2569f663c6f77ffbc15b

DESCRIPTION
```diff
@@ -30,27 +30,29 @@ Imports:
...
     utils,
     viridisLite,
     viridis,
-    rlang
+    rlang,
+    jsonlite
 Remotes:
-    bioc::rhdf5
+    bioc::rhdf5,
+    github::adokter/suntools
```
adokter: github remotes can be removed

R/utils.R
```diff
@@ -167,11 +168,232 @@ match_filenames <- function(file_list, regex_list) {
...
+
+#' Convert a dataframe into a vpts object
+#'
+#' @param data a dataframe created from a VPTS CSV file
+#' @returns a bioRad vpts object
+#' @examples
+#' df = read.csv(system.file("extdata", "example_vpts.csv", package = "bioRad"))
+#' as.vpts(df)
+#' @export
+as.vpts <- function(data) {
```
adokter, line 316: Each exported function should have its own file, so this should go into as.vpts.R

adokter (changes requested):

DESCRIPTION
adokter: fixed in [1c9492d](https://github.com/adokter/bioRad/pull/590/commits/1c9492da8569a1156c2ef4ae1c3a71500b5a8c4c)

R/read_vpts.R
```diff
@@ -1,241 +1,129 @@
...
-#'
-#' @return An object inheriting from class `vpts`, see
-#' [`vpts()`][summary.vpts] for details.
+#' Read time series of vertical profiles (`vpts`) from file(s)
 #'
+#' Reads `vpts` data from one or more files.
+#' The following file formats are supported (but cannot be mixed):
+#' - [VPTS CSV](https://aloftdata.eu/vpts-csv/).
+#' - [ODIM bird profile](https://github.com/adokter/vol2bird/wiki/ODIM-bird-profile-format-specification).
+#' - vol2bird standard output (see example below).
```
adokter, line 7: Getting a conflicting warning message:

```
> stdout_file <- system.file("extdata", "example_vpts.txt", package = "bioRad")
> 
> read_vpts(stdout_file)
No extension detected; assuming file type .txt which maps to stdout format
Error in (function (file, radar, lat, lon, height, wavelength = "C", sep = "")  : 
  'radar' argument missing. Required to specify a radar identifier.
In addition: Warning message:
In read_vpts(stdout_file) :
  .txt extenstion detected - falling back to read_stdout().

    Please consider updating your workflow by using VPTS csv or h5 input files
 ```

Comment (adokter) [reply R/read_vpts.R:7]: this file does have a .txt extension, but function says no extension detected

File: inst/extdata/example_vpts.csv
Comment (adokter) [inline inst/extdata/example_vpts.csv:1]: This file contains many duplicate timesteps, we should remove those. The package should only contain relatively 'ideal' examples, and this is clearly a flaw in the dataset for this Belgian radar

File: R/as.vpts.R
Comment (adokter) [inline R/as.vpts.R:1]: as.vpts does not act as the inverse of as.data.frame:

```
vptsfile <- system.file("extdata", "example_vpts.csv", package = "bioRad")
my_vpts <- read_vpts(vptsfile)
as.vpts(as.data.frame(my_vpts))

Error in `dplyr::mutate()` at bioRad/R/as.vpts.R:35:2:
ℹ In argument: `source_file = as.factor(source_file)`.
Caused by error:
! `source_file` must be size 3675 or 1, not 0.
Run `rlang::last_trace()` to see where the error occurred.
```

File: tests/testthat/test-read_vpts.R [hunk 23]
```diff
@@ -1,173 +1,226 @@
...
+      length(result$datetime) == 2,
+      "Expected two vp objects to be returned when reading two files."
+    )
+    # Test if the output is a vpts object
+    expect_true(is.vpts(result))
+  }
+})
+
+test_that("read_vpts() returns error on multiple radars in vp hdf5 files", {
+  skip("Not yet implemented")
```
adokter, line 112: this one is skipped, but a critical test. can we add?

adokter (changes requested):

tests/testthat/test-read_vpts.R
iskandari: read_vpts() will throws a warning instead of an error on multiple h5 radars.  This is because it invokes `bind_into_vpts()`

```r
> read_vpts(testfiles)
[[1]]
                   Regular time series of vertical profiles (class vpts)

           radar:  czbrd 
      # profiles:  3 
time range (UTC):  2023-06-01 00:00:00 - 2023-06-01 00:00:00 
   time step (s):  0 

[[2]]
                   Irregular time series of vertical profiles (class vpts)

           radar:  eehar 
      # profiles:  1 
time range (UTC):  2023-06-01 00:00:00 - 2023-06-01 00:00:00 
   time step (s):  min: NA     max:  NA 

Warning message:
In (function (...)  : Vertical profiles are not from a single radar.
```

R/as.vpts.R
iskandari: Source files are lost in the read process, fixed in [d45f23a](https://github.com/adokter/bioRad/pull/590/commits/d45f23ab2168c3755aef9ba1326887ee97864ea7)

R/read_vpts.R
iskandari: fixed in [92a8556](https://github.com/adokter/bioRad/pull/590/commits/92a855666929e386539416095a73ca2dab48068c)

inst/extdata/example_vpts.csv
iskandari: Deduped by height and datetime, fixedi in [743c70d](https://github.com/adokter/bioRad/pull/590/commits/743c70de3a065c29f0aafb0b31f54ca41c07fc5a)

adokter (approved):

Issue #562, opened 2023-05-16
warnings and errors appear in the documentation


baptischmi:
Errors and warnings appear in the *'\dontrun'* section, e.g.:

https://adriaandokter.com/bioRad/reference/example_scan

https://adriaandokter.com/bioRad/reference/example_vp

https://adriaandokter.com/bioRad/reference/example_vpts.html

iskandari:
Remove dontrun blocks that were used when Docker was still required

adokter:
closed by #598

Issue #528, opened 2022-08-31
review dontrun blocks in examples


adokter:
With vol2birdR dependency most of these blocks can be executed safely, and dontrun blocks can be removed

The aim of this review is to have the website be generated automatically in full, without the need of local runs through `pkgdown::build_site(run_dont_run=TRUE)` like we have done for CRAN releases

adokter:
See also #562, the examples listed in #562 should remain in dontrun blocks, so they are kept out of the documentation.

adokter:
closed by #598

PR #597, opened 2023-06-07
review dontrun blocks in examples (#528)

adokter:
Addressing issue #528 

Removed dontrun blocks as much as possible. Remaining example blocks in \dontrun blocks are there because:
* execution time > 5s, which could trigger a note on CRAN
* the example downloads remote data, and therefore requires an internet connection

Also made sure a `devtools::check(run_dont_run=TRUE)` completes successfully without triggering warnings/errors, fixing #562


adokter:
closing PR because of a github error, starting over fresh

PR #601, opened 2023-06-08
Re-enable tests for read_stdout()

peterdesmet:
Tests for `read_vpts()` were added by @PietrH. As part of #590 that function is renamed to `read_stdout()`, causing a merge conflict with between the tests for the new `read_vpts()` and those intended for `read_stdout()`. This resulted in the latter test being removed in https://github.com/adokter/bioRad/commit/3358d976e39fefb6e1d8eac80fdf1b93faf9665e.

This PR re-enables those tests, now calling `read_stdout()`. I notice however that some of the tests fail. I think this is due to the fact the tests expecting a specific error are order sensitive. E.g. a test for `read_stdout(vptsfile, lat = "a")` expects an error related to the latitude not being an integer, but can also fail because the radar is missing. I would update those test to provide all the necessary parameters _except_ the one we need to test, so `read_stdout(vptsfile, radar = "KGBM", lat = "a")`.

@PietrH since you worked on these test, would you be willing to correct them?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: bioRad
 Title: Biological Analysis and Visualization of Weather Radar Data
-Version: 0.7.0.9590
+Version: 0.7.0.9601
 Description: Extract, visualize and summarize aerial movements of birds and
     insects from weather radar data. See <doi:10.1111/ecog.04028>
     for a software paper describing package and methodologies.
--- a/R/hooks.R
+++ b/R/hooks.R
@@ -1,8 +1,5 @@
 .onLoad <- function(libname, pkgname) {
   register_all_s3_methods() # dynamically registers non-imported pkgs (tidyverse) # nocov
-  if (inherits(try(find.package("vol2birdR"), silent = TRUE),"try-error")) {
-    warning("Package 'vol2birdR' is not found. Please see https://adokter.github.io/vol2birdR/ for installation instructions.")
-  }
 }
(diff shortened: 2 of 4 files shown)


tests/testthat/helper.R
```diff
@@ -0,0 +1,43 @@
...
+  dest_file <- file.path(dest_sub_dir, file_name)
+
+  # Download the file if it doesn't already exist
+  if (!file.exists(dest_file)) {
+    message("Downloading ", url)
+    curl::curl_download(url, destfile = dest_file)
+  } 
+}
+
+sapply(urls, download_test_file)
```
PietrH, line 43: Could we not save these in a temp dir, and store the path in test environment instead of storing these test files in test root. 

I also think the call of the helper should be in the test file that is using the files. That way we can tell if the test files ever become unused.

PietrH: If you run `devtools::load_all()` you'll notice the files in your project root.

PietrH:
@peterdesmet , I believe I have resolved all testing woes. 

Could you have a look at `tests/testthat/helper.R` before merging? I believe a helper should not change the testing environment.

tests/testthat/helper.R
peterdesmet: I agree and have noted this in my review to @iskandari  of https://github.com/adokter/bioRad/pull/590#discussion_r1219928093

peterdesmet:
I have changed the base for this PR to #590. The helper function can be tackled there.

PR #626, opened 2023-07-26
handle delta linear interpolation if it is a numeric of len 0

iskandari:
Resolves error that arises when `delta_linear_interpolation` is numeric(0)

```
Error in if (is.na(delta_linear_interpolation)) delta_linear_interpolation = 0 : 
  argument is of length zero
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: bioRad
 Title: Biological Analysis and Visualization of Weather Radar Data
-Version: 0.7.1.9000
+Version: 0.7.1.9001
 Description: Extract, visualize and summarize aerial movements of birds and
     insects from weather radar data. See Dokter, A. M. et al. (2018) 
     "bioRad: biological analysis and visualization of weather radar data" <doi:10.1111/ecog.04028>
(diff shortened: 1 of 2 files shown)


DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: bioRad
 Title: Biological Analysis and Visualization of Weather Radar Data
-Version: 0.7.1.9000
+Version: 0.7.1.9001
```
adokter, line 3: We decided to stop increasing dev version and have more frequent official releases. So can keep this at 9000

R/integrate_profile.R
```diff
@@ -262,7 +262,9 @@ integrate_profile.vp <- function(x, alt_min = 0, alt_max = Inf, alpha = NA,
     height_quantile_upper <- denscum[min(height_index_lower+1,length(denscum))]
     # 4) do a linear interpolation to estimate the altitude at the quantile of interest
     delta_linear_interpolation <- (height_quantile-height_quantile_lower)*(height_upper-height_lower)/(height_quantile_upper-height_quantile_lower)
-    if(is.na(delta_linear_interpolation)) delta_linear_interpolation=0
+    if(length(delta_linear_interpolation) == 0 || is.na(delta_linear_interpolation)) {
```
adokter, line 265: Do we understand how this can become NA, is that outcome expected?

adokter (changes requested):

adokter:
closing and continuing fix in #627

PR #631, opened 2023-09-01
Bugfix/cran 0.7.1

iskandari:
- Radar reference changed to [JSON](https://raw.githubusercontent.com/enram/aloftdata.eu/main/_data/OPERA_RADARS_DB.json) 
- Makes tests compatible with `testthat 3.0.0`
- Addresses CRAN warnings


DESCRIPTION
```diff
@@ -30,37 +30,37 @@ Depends:
...
     ggmap (>= 3.0.0),
     knitr,
-    tidyselect,
     rmarkdown,
+    sf,
     testthat (>= 3.0.0),
+    tidyselect,
     vdiffr,
-    sf
+    vol2birdR
```
adokter, line 63: Several collaborators wanted to keep this as a Suggest, and checks were added to all functions that use vol2birdR that prompt the user to install it when needed. Can't we keep it like that?

tests/testthat/test-pvolfile.R
```diff
@@ -9,8 +9,14 @@ ghostfile <- paste(tempdir(), "garbage", sep = "/")
 test_that("is.pvolfile() can identify ODIM HDF5 polar volume files", {
   expect_true(is.pvolfile(pvolfile))
   expect_false(is.pvolfile(vpfile))
-  expect_warning(expect_false(is.pvolfile(textfile)))
-  expect_warning(is.pvolfile(ghostfile), "Can't find") # From get_odim_object_type
+
+  is.pvolfile(textfile) %>%
```
adokter, line 13: If I point `textfile to a non-text file I get this, I don't think you can pipe one expect_warning() into another:
```
is.pvolfile(textfile) %>%
+   expect_warning("is not an HDF5 file") %>%
+   expect_warning("other data formats not yet recognized")
Error: `.` did not produce any warnings.
```

adokter (changes requested):

iskandari:
@adokter looks like [ubuntu-latest (devel)] (https://github.com/adokter/bioRad/actions/runs/6053457936/job/16428992890#logs)
build always times out, otherwise checks pass

DESCRIPTION
iskandari: Yes it seems so

tests/testthat/test-pvolfile.R
iskandari: Seem like you can on the latest version of `testthat` and it is also what the package maintainers now recommend in [the docs](https://testthat.r-lib.org/articles/third-edition.html)

```r
> is.pvolfile(ghostfile) %>%
+     expect_warning("Can't find") %>%
+     expect_warning("other data formats not yet recognized")
> packageVersion("testthat")
[1] ‘3.1.8’
```

iskandari: I think in the latest versions you can, but the function def still has bugs.  For example, these work:

```r
> packageVersion("testthat")
[1] ‘3.1.8’
> is.pvolfile(textfile) %>% expect_warning("is not an HDF5 file") %>% expect_warning('other data formats')
> is.pvolfile(ghostfile) %>%
+     expect_warning("Can't find") %>%
+     expect_warning("other data formats not yet recognized")
```

adokter (approved):

PR #674, opened 2024-09-16
fix label of plot.vpi() for quantity vir, disable read_vpts test

adokter:
* fixes a mislabeled plot axis for vertically integrated reflectivity (VIR). y-axis label incorrectly mentioned cm^2/km/h, while it should be cm^2/km^2
* disables a test in test-read_vpts.R related to an issue in `read_csv()` of package readr


fyiguo (approved):
the label units are now correct

Issue #682, opened 2024-10-31
read_pvolfile() warning "NAs produced by integer overflow"


adokter:
For certain Canadian ODIM h5 files we get warning
```
1: In H5Aread(A, ...) :
  NAs produced by integer overflow while converting 64-bit integer from HDF5 to a 32-bit integer in R.
Choose bit64conversion='bit64' or bit64conversion='double' to avoid data loss I am not sure I totally understand
```
We call H5Aread indirectly by use of `rhdf5::h5readAttributes()` (which accepts bit64conversion as an argument) in functions `bioRad::read_pvolfile()` and `bioRad::read_vpfiles()` and `bioRad::get_odim_object_type()`

Need to find out which attribute throws this warning, and whether we need to change default behavior. Argument `bit64conversion='bit64'` adds dependency on bit64 package, and `bit64conversion='double'` turns integers into floats, both not ideal. `bit64conversion='int'` coercing to 32-bit integers is enforced, with the risk of data loss, but with the insurance that numbers are represented as integers (see rhdf5::H5Aread() details section)

adokter:
The warning is thrown for attributes `pvol$scans[[i]]$attributes$how$noisepwerh` and `pvol$scans[[i]]$attributes$how$noisepwerv` for all scans `i`

adokter:
noisepowerh and noisepowerv are defined as 32-bit numbers, but stored as 64 Bit datatype H5T_STD_I64LE in the hdf5 file

adokter:
attributes noisepowerh and noisepowerv not documented in ODIM H5 2.4.1 (July 2024) https://www.eumetnet.eu/wp-content/uploads/2024/10/ODIM_H5_v2.4.1_final.pdf

adokter:
fixed by https://github.com/adokter/bioRad/commit/f50428ec65e827792103e2442d3a09d6236efa49

adokter:
Fix https://github.com/adokter/bioRad/commit/f50428ec65e827792103e2442d3a09d6236efa49 introduces an issue with reading hdf5 files that have been written by write_pvolfile() - holding off on a fix for now

PR #683, opened 2024-11-01
fix 64 bit read error for attributes (#682)

adokter:
R doesn't support 64 bit integers. This PR casts 64 bit integer attributes to doubles such that they can be read by R


adokter:
This PR breaks the ODIM h5 format after writing a file with write_pvolfile(). Resulting hdf5 file cannot be read by RaveIO_open() function in Rave.

PR #693, opened 2025-01-14
Plotting tweaks: bug fix for x label, and support night shading with missing data

dsheldon:
Change 1: fix a bug where the `xlab` argument passed to `plot.vpi` is not used and the x label is hard coded to "time" when `plot` is called.

Change 2: ignore NAs when automatically determining the y axis limits for night shading. This allows the night shading to work properly if the plotted quantity has NAs, which can be useful if you want to filter some values from plotting.


adokter:
Thanks @dsheldon!

PR #698, opened 2025-02-26
bird insect partitioning and post-hoc precipitation filter

### LINKED ISSUE #697: Duplicated timestamps in csv change height distribution in vpts plot
While exploring some data I noticed the height distribution of some `vpts` plots looking strange. Here somehow the data get plotted 2 or 3 times as high. It might related to the `height` `datetime` combination being replicated in the source file. The lower height distribution seems correct see also (https://crow.aloftdata.eu#/?radar=fikor&date=2024-10-12&interval=72&timedisplay=radarLocal&vpiMode=mtr&vpColorScheme=viridis&lang=en):

``` r
require(bioRad)
#> Welcome to bioRad version 0.8.1
#> using vol2birdR version 1.0.5 (MistNet not installed)
require(lubridate)
#> Loading required package: lubridate
#> 
#> Attaching package: 'lubridate'
#> 
#>     date, intersect, setdiff, union
t<-as.Date("2024-10-12")+lubridate::hours(15)+lubridate::minutes(0:(12*8)*5)
int<-as.interval(min(t), max(t))
# r<-bioRad::download_vpfiles(as.character(min(as.Date(t))),
#                             as.character(max(as.Date(t))), "fikor", directory = "/home/bart/")
vptsAloft<-read_vpts("/home/bart/fikor_vpts_202410.csv.gz")
vptsAloft<-vptsAloft[vptsAloft$datetime %within% int]
plot(vptsAloft)
#> Warning in plot.vpts(vptsAloft): Irregular time-series: missing profiles will
#> not be visible. Use 'regularize_vpts' to make time series regular.
#> Warning in plot.vpts(vptsAloft): Dropped 66 profiles with duplicate datetime
#> values
```

``` r
vptsAloft<-vptsAloft[!duplicated(vptsAloft$datetime)]
plot(vptsAloft, main="Dups removed")
```

``` r
require(dplyr)
#> 
#> Attaching package: 'dplyr'
#> 
#>     filter, lag
#> 
vroom::vroom('/home/bart/fikor_vpts_202410.csv.gz')  |>
  filter(datetime %within%int) |>
  select(dens, height, datetime) |> group_by(datetime, height) |>
  summarise(dens=tail(head(dens,1),1))|> ungroup()|>
  tidyr::pivot_wider(names_from = "height", values_from = "dens") |>
  select(-datetime) |> as.matrix() |> image()
#> Rows: 200125 Columns: 26
#> ── Column specification ────────────────────────────────────────────────────────
#> Delimiter: ","
#> chr   (2): radar, source_file
#> dbl  (21): height, u, v, w, ff, dd, sd_vvp, eta, dens, dbz, dbz_all, n, n_db...
#> lgl   (2): gap, vcp
#> dttm  (1): datetime
#> 
#> ℹ Use `spec()` to retrieve the full column specification for this data.
#> ℹ Specify the column types or set `show_col_types = FALSE` to quiet this message.
#> `summarise()` has grouped output by 'datetime'. You can override using the `.groups` argument.
```

``` r

vroom::vroom('/home/bart/fikor_vpts_202410.csv.gz')  |>
  filter(datetime %within%int) |> group_by(height)|>
  filter(!duplicated(datetime)) |> as.data.frame()|> bioRad:::as.vpts() |> plot()
#> Rows: 200125 Columns: 26
#> ── Column specification ────────────────────────────────────────────────────────
#> Delimiter: ","
#> chr   (2): radar, source_file
#> dbl  (21): height, u, v, w, ff, dd, sd_vvp, eta, dens, dbz, dbz_all, n, n_db...
#> lgl   (2): gap, vcp
#> dttm  (1): datetime
#> 
#> ℹ Use `spec()` to retrieve the full column specification for this data.
#> ℹ Specify the column types or set `show_col_types = FALSE` to quiet this message.
```

<sup>Created on 2025-02-25 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

adokter:
- Implements #700 
- implements #701
- fixes #697


R/as.vp.R
```diff
@@ -0,0 +1,19 @@
...
+#' df <- as.data.frame(example_vp)
+#' # convert the data.frame to a vp object:
+#' as.vp(df)
+#' @export
+as.vp <- function(data) {
+  assertthat::assert_that(inherits(data,"data.frame"))
+
+  vpts <- as.vpts(data)
+  
+  assertthat::assert_that(length(vpts$datetime) == 1, msg="multiple timestamps found, data is not a single vertical profile")
```
bart1, line 16: You might want to extent this check to if height bins are duplicated. Also do height bins need to be regularly spaced?

adokter: I believe those checks are performed already by as.vpts() which is used under the hood, but I'll double check

adokter: Rows can be removed, so no check for regular spacing currently:
> df=as.data.frame(example_vp, suntime=F)
> as.vp(df[-5,])

adokter: related to https://github.com/adokter/bioRad/issues/697, which requires a fix before next release

adokter:
@bart1 can you please review if this PR fixes your bug report on duplicates timestamps (#697)

bart1:
Ill check

Sent from Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

R/clean_mixture.R
```diff
@@ -0,0 +1,252 @@
...
+#' single-polarization weather radar data. Remote Sensing. 2021 May 19;13(10):1989 \doi{10.3390/rs13101989}.
+#' @details
+#' For a detail description of the methodology see Shi et al. (2025).
+#' Most commonly the fast component refers to migrating birds, while
+#' the slow component refers to insects. The slow component is always
+#' oriented in the direction of the wind by definition.
+#' Note that for mixture airspeeds exceeding the airspeed of the fast component,
+#' all reflectivity is assigned to the fast component. Similarly, for mixture
+#' airspeeds below the airspeed of the slow component, all reflectivity
+#' will be assigned to the slow component. In both these limiting cases the
```
XuShi94, line 76: comment unfinished here

bart1:
@adokter Sorry for the delay, I just start working a bit with this, somethings I noticed below

First of all nice changes, useful to make script easier! On the topic of correcting VPTS data, maybe we could also think about filling holes in a vpts when the `sdvvp` drops to low during intense migration. That is a other frequent fix I used. 

# not equally high profiles

I noticed `as.vpts` requires the height profiles to be exactly equal otherwise it drops the profiles. If the height interval is equal (e.g. 200 meters) it might be nice to just `NA` fill the profiles? I think the aloft profiles for example not always go up to the same height:
``` r
devtools::load_all("~/bioRad/")
#> ℹ Loading bioRad
#> Welcome to bioRad version 0.8.1.9000
#> 
#> using vol2birdR version 1.0.5 (MistNet not installed)
pvolfile_source <- system.file("extdata", "volume.h5", package = "bioRad")

pvolfile <- paste0(tempdir(),"/volume.h5")
file.copy(pvolfile_source, pvolfile)
#> [1] TRUE

  vp <- calculate_vp(pvolfile) |> as.data.frame()
#> Running vol2birdSetUp
#> Warning: radial velocities will be dealiased...
  vp_low <- calculate_vp(pvolfile, n_layer = 10)|> as.data.frame()
#> Running vol2birdSetUp
#> Warning: radial velocities will be dealiased...
  
as.vpts(rbind(vp, vp))
#> Warning in validate_vpts(data): Extra fields found: day, sunrise, sunset
#>                    Regular time series of vertical profiles (class vpts)
#> 
#>            radar:  seang 
#>       # profiles:  2 
#> time range (UTC):  2015-10-18 18:00:00 - 2015-10-18 18:00:00 
#>    time step (s):  0
as.vpts(rbind(vp, vp, vp_low))
#> Warning in validate_vpts(data): Extra fields found: day, sunrise, sunset
#> Warning in as.vpts(rbind(vp, vp, vp_low)): Profiles found with different altitude layer specifications, keeping only the most
#>     common one, dropping  1 profile(s) ...
#> Warning in as.vpts(rbind(vp, vp, vp_low)): Profiles found with different number
#> of height layers: 10 20, retaining 20 only.
#>                    Regular time series of vertical profiles (class vpts)
#> 
#>            radar:  seang 
#>       # profiles:  2 
#> time range (UTC):  2015-10-18 18:00:00 - 2015-10-18 18:00:00 
#>    time step (s):  0
```

Here it would be nice if `vp_low` was just filled with NA values especially if the difference is in the high range (e.g. a mixture of profiles upto 4000 and 5000 meter)

# Clean mixture

I would watch out with the argument naming `u` vs `U`, and `v` vs `V`, would it not be better to use something like `V_wind`. I know most of the time it won't be needed but still I could see bugs  slipping in easily. I find capitalization to distinguish parameters difficult.

# Filter precip

Maybe you can use something like this as a default parameter for dbz : `ifelse(x$attributes$how$wavelength<7,x, y)`. In this way you can optimize the default for both C and S band, that should at least help with the last note in the help description.

# #697 

Is mostly resolved except the `source_file` seems to be mixed up. Not a big issue as I expect not many people to use it but still it would be good to make sure that is correct

adokter:
Thanks @bart1, I'll start working on incorporating this feedback

PR #739, opened 2025-09-02
Adding velocity azimuth display plots

bart1:
In this pull request i want to introduce a VAD plotting function. It helps to explain and asses the fit of velocities in a vp. There is considerable flexibility in formatting the plots. Here are some example plots:

``` r
devtools::load_all("~/bioRad")
#> ℹ Loading bioRad
#> Welcome to bioRad version 0.11.0.9000
#> 
#> using vol2birdR version 1.1.1 (MistNet not installed)
pvolfile <- system.file("extdata", "volume.h5", package = "bioRad")
example_pvol <- read_pvolfile(pvolfile)
vad(example_pvol,
  range = c(5000, 30000), height = c(200, 400),
  point_geom_args = list(size = 0.5)
)
```

``` r
vp <- calculate_vp(pvolfile)
#> Running vol2birdSetUp
#> Warning: radial velocities will be dealiased...
vad(example_pvol,
  vp = vp, height = c(400, 1200),
  point_geom = ggplot2::geom_bin2d,
  annotate = "sdvvp: {round(sd_vvp,2)} [m/s]",
) + ggplot2::scale_fill_viridis_c()
```


adokter:
Thanks @bart1, this will be helpful.

R/vad.R
```diff
@@ -0,0 +1,183 @@
...
+#'   vp = vp, height=c(400,1200),
+#'   plotting_geom=ggplot2::geom_bin2d,
+#'   annotate="sdvvp: {round(sd_vvp,2)} [m/s]",
+#' ) + ggplot2::scale_fill_viridis_c()
+vad <- function(x, ...) {
+  UseMethod("vad", x)
+}
+#' @rdname vad
+#' @export
+vad.pvol <- function(x, vp = NULL,..., range = NULL, height = NULL,
```
adokter, line 63: In general there are very little inline comments, it would be helpful to have more text added to understand what happens in each code block in the function

R/vad.R
```diff
@@ -0,0 +1,183 @@
+globalVariables(c("DBZH","VRADH","RHOHV"))
```
adokter, line 1: why do we need this?.

R/vad.R
```diff
@@ -0,0 +1,183 @@
...
+#'   plotting_geom=ggplot2::geom_bin2d,
+#'   annotate="sdvvp: {round(sd_vvp,2)} [m/s]",
+#' ) + ggplot2::scale_fill_viridis_c()
+vad <- function(x, ...) {
+  UseMethod("vad", x)
+}
+#' @rdname vad
+#' @export
+vad.pvol <- function(x, vp = NULL,..., range = NULL, height = NULL,
+                     range_gate_filter = DBZH < 20 & RHOHV < .95,
```
adokter, line 64: Generally there are many flavors of radial velocity and reflectivity allowed in the ODIM specification. So the parameter names you want to plot should be parameters to the function. For radial velocity I have seen VRAD, VRADH, VRADV, for reflectivity we have DBZH, DBZV, TH, TV, so better to keep flexibility

R/vad.R
```diff
@@ -0,0 +1,183 @@
...
+    # )
+    vp_geom <- mapply(SIMPLIFY = F,
+      function(spd, dir,hgt, bin)
+      {ggplot2::geom_function(
+        data = dplyr::bind_cols(data, data.frame(
+          min_bin_height = hgt,
+          max_bin_height = hgt + vp$attributes$where$interval,
+          height_bin = bin
+        )),
+        fun = function(x, v, a) cos((x - a) / 180 * pi) * v,
```
adokter, line 146: There is an issue with wanting to plot data from multiple elevation scans using the same formula. Because the radar measures radial velocity, there should be `cos(elevation_angle)` factor in this plotting function. E.g. for the extreme case where the radar looks vertically up, you will have radial velocities all zero for any speed and direction. Therefore you cannot just plot the same curve to data from different elevation scans.

It seems more natural for this function to not act on a polar volume, but on a scan.

If you really want a method that plots all data for a volume, you might be able to "correct" the raw radial velocity data for the reduction in radial velocity due to the elevation angle. Then you might be able to plot multiple scans in one graph - but we would have to very carefully document what is actually happening and that the user is no longer looking at pristine raw data. For a 20 degree elevation angle scan you are looking at a correction of more than 5%

R/vad.R
```diff
@@ -0,0 +1,183 @@
...
+  if (is.null(range)) {
+    range <- c(-Inf, Inf)
+  }
+  if (is.null(height)) {
+    height <- c(-Inf, Inf)
+  }
+  data <-
+    mapply(SIMPLIFY = F,
+      cbind,
+      lapply( lapply(x$scans, scan_to_spatial),
```
adokter, line 109: I think you need an option to select an elevation scan range. For example, for a 90 degree bird bath scan this won't work. See also my comment below on the effect of elevation angle on detected radial velocity

R/vad.R
adokter: This is the function we fit the radial velocity field to, with theta elevation angle, phi azimuth, and u,v the components of the horizontal speed and w the vertical speed:
VRAD = w sin(theta) + u cos(theta) sin(phi) + v cos(theta) cos(phi)
For birds you can usually assume w is zero (but we do estimate it), so it's just a simple extra cos(theta) factor that needs to be added when plotting a scan. Or alternatively, when aligning multiple scan into one plot, the VRAD values needed to be divided by cos(theta)

R/vad.R
```diff
@@ -0,0 +1,183 @@
...
+#' deviations from the sine function.
+#'
+#' @param x A polar volume from which range gates are extracted.
+#' @param ... Currently not used.
+#' @param vp A vertical profile to annotate the VAD plot with the fit in the vertical profile
+#' @param range The distance range in to filter the range gates with.
+#'      If a `vp` is provided the range is taken from there and the argument `range` should not be provided.
+#' @param height The height range to filter the range gates by.
+#'      If only one value is provided next to a `vp` then the height bin intersection with this elevation is plotted.
+#' @param range_gate_filter Optional filtering of the range gates. By default range gates are filtered for a `DBZH` less then 20 and a `RHOHV` less then 0.95.
```
adokter, line 18: please use reflectivity filtering on the linear eta scale (see dbz_to_eta()) - vol2bird currently uses 36000 as the cutoff. This has the advantage that effectively the same threshold is used at C and S-band

adokter: or alternatively the default should look at whether the radar is S- or C-band and adjust the default threshold, 20 dBZ is too low for S-band

R/vad.R
```diff
@@ -0,0 +1,183 @@
...
+
+#' Create a Velocity Azimuth Display (VAD) plot
+#'
+#' A velocity azimuth display plot visualized the radial velocity (`VRAD`) as a function of the azimuth from the radar.
+#' Among others it can be used to asses the fit of the movement speeds in a vertical profile.
+#' For example, when the movement is not uniform but rather in multiple directions or rotating this will be visible as
+#' deviations from the sine function.
+#'
+#' @param x A polar volume from which range gates are extracted.
+#' @param ... Currently not used.
```
adokter, line 12: why do we need it if it is not used?

R/vad.R
```diff
@@ -0,0 +1,183 @@
...
+#' @param vp A vertical profile to annotate the VAD plot with the fit in the vertical profile
+#' @param range The distance range in to filter the range gates with.
```
adokter, line 14: other functions in bioRad use `range_min` and `range_max`, so I suggest using the same naming convention. `range` is more used for the continuous variable of radar range, not to set upper/lower limits

R/vad.R
```diff
@@ -0,0 +1,183 @@
...
+#'      If a `vp` is provided the range is taken from there and the argument `range` should not be provided.
+#' @param height The height range to filter the range gates by.
```
adokter, line 16: other functions use `alt_min` and `alt_max` to specify height ranges, suggest using the same nomenclature

R/vad.R
```diff
@@ -0,0 +1,183 @@
...
+#'      If only one value is provided next to a `vp` then the height bin intersection with this elevation is plotted.
+#' @param range_gate_filter Optional filtering of the range gates. By default range gates are filtered for a `DBZH` less then 20 and a `RHOHV` less then 0.95.
+#'      Alternative filters could be used to highlight specific effects.
+#' @param plotting_geom The geom function to visualize the range gates, the default is [ggplot2::geom_point()], in some cases this suffers from over plotting.
+#'      Alternatives that avoid over plotting could be [ggplot2::geom_bin2d()] or [ggpointdensity::geom_pointdensity()].
+#' @param plotting_geom_args A list with additional arguments to the `plotting_geom` function. For example, controlling the point size or alpha.
+#' @param annotate A [glue][glue::glue()] string that is used to annotate the plot with additional properties of the height bin of the `vp`.
+#'      The string is evaluated using the columns from `as.data.frame(vp)`.
+#'      Use `NULL` if no annotation is desired.
+#' @param vp_color The color used for the `vp` annotations and line.
```
adokter, line 26: rename to `annotation_color`? This goes together with `annotation_size` right?

R/vad.R
```diff
@@ -0,0 +1,183 @@
...
+#'      If a `vp` is provided the range is taken from there and the argument `range` should not be provided.
+#' @param height The height range to filter the range gates by.
+#'      If only one value is provided next to a `vp` then the height bin intersection with this elevation is plotted.
+#' @param range_gate_filter Optional filtering of the range gates. By default range gates are filtered for a `DBZH` less then 20 and a `RHOHV` less then 0.95.
+#'      Alternative filters could be used to highlight specific effects.
+#' @param plotting_geom The geom function to visualize the range gates, the default is [ggplot2::geom_point()], in some cases this suffers from over plotting.
+#'      Alternatives that avoid over plotting could be [ggplot2::geom_bin2d()] or [ggpointdensity::geom_pointdensity()].
+#' @param plotting_geom_args A list with additional arguments to the `plotting_geom` function. For example, controlling the point size or alpha.
+#' @param annotate A [glue][glue::glue()] string that is used to annotate the plot with additional properties of the height bin of the `vp`.
+#'      The string is evaluated using the columns from `as.data.frame(vp)`.
```
adokter, line 24: ? clarify - hard to understand what this means

adokter (changes requested):

R/vad.R
bart1: It now uses  the quantities present, first the single polarization version then horizontal and if both are not present the vertical.

R/vad.R
bart1: changed

R/vad.R
adokter: ok - we don't use this in other functions. Could you add an issue to discuss if we should use this consistently throughout the package, so others can weigh in as well

R/vad.R
```diff
@@ -0,0 +1,263 @@
...
+#' @param vp An object of class `pvol`, typically a vertical profile estimated for the input polar volume specified under `x` using `bioRad::calculate_vp()`.
+#' @param ... Currently not used.
+#' @param range_min,range_max Numeric. The minimum and maximum range to include, in m. Values are taken from the `vp`
+#' object if provided.
+#' @param alt_min,alt_max Numeric. The minimum and maximum altitude to include, in m. If only `alt_min` value is provided
+#' next to a `vp` then the height bin intersection with this altitude is plotted.
+#' @param range_gate_filter Optional filtering of the range gates. By default range gates are filtered for linear reflectivity
+#' value (`eta`) less then 36000 cm^2/km^3 (the default in the profiling algorithm, see `etaMax` in `vol2bird::vol2bird_config()`)
+#'  and a correlation coefficient (`RHOHV`) < 0.95. The function selects the first reflectivity factor quantity
+#'   from `DBZ`, `DBZH`, `DBZV`, `TH` or `TV` that is present. Alternative filters could be used to highlight specific effects.
```
adokter, line 20: It would be better if the user could specify which reflectivity quantity to use, instead of defaulting to one.

R/vad.R
```diff
@@ -0,0 +1,263 @@
...
+#' See details for more information on the specific options.
+#'
+#' @returns A [ggplot2::ggplot] object.
+#'
+#' @details The returned [ggplot2::ggplot] can be styled with additional elements available in the `ggplot2` package.
+#' For example, labels and titles can be modified using [ggplot2::labs()] or [ggplot2::ggtitle()].
+#' Using [ggplot2::theme()] the visual appearance can easily be modified. Use the regular [+][ggplot2::+.gg]
+#' syntax to make modifications (e.g. see the examples for changing the plotting color gradients).
+#'
+#' As for the radial velocity to plot the first of `VRAD`, `VRADH` or `VRADV` is used.
```
adokter, line 44: better to give the user the flexibility to select which quantity to use

R/vad.R
```diff
@@ -0,0 +1,263 @@
...
+#' @param plotting_geom The geom function to visualize the range gates, the default is [ggplot2::geom_point()]. In cases
+#' with many data points plots may become cluttered, in which cases alternatives like [ggplot2::geom_bin2d()] or
+#'  [ggpointdensity::geom_pointdensity()] may be preferred.
+#' @param plotting_geom_args A list with additional arguments to the `plotting_geom` function. For example, controlling the point size or transparancy alpha.
+#' @param annotate A [glue][glue::glue()] formating string that is used to annotate the plot with additional properties of the height bin from the `vp`.
+#' The string is evaluated using the columns from `as.data.frame(vp)`, any of these columns can thus be used (e.g. `ff` or `sd_vvp`).
+#' Use `NULL` if no annotation is desired.
+#' @param annotation_color The color used for the `vp` annotations and line.
+#' @param annotation_size The text size used for the annotation.
+#' @param cosine_correction A character option to select what approach should be taken to correct for the fact
```
adokter, line 30: I think the word `cosine_correction` is misleading - there is no ture correction, the cosine is simply part of the model fit, so I think we shouldn't confuse the reader with there being a correction applied. We should just plot the fit and the datapoints, nothing more.

The main sticking point remains how to plot multiple sweeps into one plot. You either need to adjust the positions of the points to "absorb" the part of the model variation with elevation angle (what you call "range_gates"), so you can keep one fitting line, or - alternatively - plot the points as they are but include multiple fit lines for each sweep (coloring points and fitting lines by sweep number).

To me the more natural user option would be to be able to toggle between those two. The user argument would then by `align_scans = TRUE/FALSE`, with the FALSE case plotting multiple fit lines (not requiring any adjustments of the points), and the TRUE case adjusting the points and showing only one line.

R/vad.R
```diff
@@ -0,0 +1,263 @@
...
+#' ) + ggplot2::scale_fill_viridis_c()
+vad <- function(x, ...) {
+  UseMethod("vad", x)
+}
+#' @rdname vad
+#' @export
+vad.pvol <- function(x, vp = NULL, ...,
+                     range_min = NULL, range_max = NULL,
+                     alt_min = NULL, alt_max = NULL,
+                     range_gate_filter =
```
adokter: I think this argument is overly complicated. Why not give single arguments `rho_hv` (same name used in `calculate_vp()` and `eta_max` to set the two thresholds. I realize this gives you less flexibility, but I think it's better to keep things simple for the user.

R/vad.R
```diff
@@ -0,0 +1,263 @@
...
+  vp_geom <- list()
+  if (!is.null(vp)) {
+    # restrict the vp annotation to only those elevation height bins for which we have data
+    vp_df <- dplyr::filter(vp_df, !(height + vp$attributes$where$interval < min(data$height) | height > max(data$height)))
+
+    elangles <- unique(data$where.elangle)
+    if (length(elangles) != 1 && cosine_correction == "vp") {
+      warning("The data to plot is based on multiple elevation angles. To correct the `vp` data one elevation angle is needed, it is therefore averaged resulting in a inperfect correction.")
+    }
+    mean_elangle_cos <- cospi(mean(elangles) / 180) # TODO should this be a weighted mean?
```
adokter, line 220: I think instead of the mean, points should be corrected according to the elevation scan they belong to, so different cosine corrections for each of the elevation scans of the volume

R/vad.R
```diff
@@ -0,0 +1,263 @@
...
+  }
+  ylab_label <- "Radial velocity [m/s]"
+  if (cosine_correction == "range_gates") {
+    data[, vrad_quantity] <- data[, vrad_quantity] * (1 / cospi(data$where.elangle / 180))
+    ylab_label <- "Corrected radial velocity [m/s]"
+    if (min(data$how.NI) < 25) {
+      warning("There are data that have a relatively low nyquist velocity (below 25 m/s) meaning nyquist folding is likely to occur. Applying a cosine correction to range gates that are nyquist folded can result in larger deviations then before.")
+    }
+  }
+  if (cosine_correction == "none" && max(data$where.elangle) > acos(.95) / pi * 180) {
```
adokter: why acos(.95) / pi * 180?

R/vad.R
```diff
@@ -0,0 +1,263 @@
...
+                         utils::head(dplyr::matches(c("^DBZ$", "^DBZH$", "^DBZV$", "^TH$", "^TV$")), 1),
+                         \(dbz) dbz_to_eta(dbz, !!x$attributes$how$wavelength) < 36000
+                       ) &
+                         dplyr::if_any(dplyr::matches("RHOHV"), \(rhohv) rhohv < .95),
+                     plotting_geom = ggplot2::geom_point,
+                     plotting_geom_args = list(),
+                     annotate = "{round(ff,1)} m/s, {round(dd)}\u00B0",
+                     annotation_size = 4,
+                     annotation_color = "red",
+                     cosine_correction = c("none", "vp", "range_gates")) {
```
adokter: To keep it simple for the user, why don't we use what you now have under "vp" when plotting a single scan (so taking account the cosine in the plotted sine curve), and when you plot a "pvol" what you now have under "range_gates". Alternatively, you could plot multiple sine curves in different colors, and highlight points in the same colors to indicate which scan they belong too.

adokter (changes requested):
Not a fully finished review, but these are the most important points, let me know what you think

R/vad.R
```diff
@@ -0,0 +1,263 @@
...
+#' @returns A [ggplot2::ggplot] object.
+#'
+#' @details The returned [ggplot2::ggplot] can be styled with additional elements available in the `ggplot2` package.
+#' For example, labels and titles can be modified using [ggplot2::labs()] or [ggplot2::ggtitle()].
+#' Using [ggplot2::theme()] the visual appearance can easily be modified. Use the regular [+][ggplot2::+.gg]
+#' syntax to make modifications (e.g. see the examples for changing the plotting color gradients).
+#'
+#' As for the radial velocity to plot the first of `VRAD`, `VRADH` or `VRADV` is used.
+#'
+#' For these plots no vertical velocity is assumed as aeroecology will predominantly move in the horizontal plane.
```
adokter, line 46: I think we *should* take into account the fitted vertical velocity when plotting the model fit from the vp, it is part of the fit and easy to incorporate - available as quantity `w` in the profile. The full radial velocity functional form is 
```
VRAD = w sin(theta) + u cos(theta) sin(phi) + v cos(theta) cos(phi)
```
so you only need to add a term `w sin(theta)`

R/vad.R
```diff
@@ -0,0 +1,263 @@
+#' Plot a Velocity Azimuth Display (VAD).
```
adokter, line 1: This high-level comment I should have raised earlier: 
- the profile fits of velocity and direction are altitude-specific. So therefore you cannot show multiple altitude bins in one plot, it then becomes unclear which speed vector to use for the model fit. How does the current implementation deal with this?

If you really want to show multiple altitude layers with different speed vectors in one plot, I feel we should show the sine curve for each altitude, and label the points for each layer in a different color.

There can be very strong altitudinal shifts in direction, easily up to 180 degrees

adokter: never mind, I see you are plotting already separate plots for each altitude layer

R/vad.R
```diff
@@ -0,0 +1,263 @@
...
+#' and for visually inspecting the quality of the velocity fit of a vertical profile estimate.
+#'
+#' Poor velocity fits in vertical profiles can arise in cases where the movement is not well described by a unidirectional
+#' velocity model, or for data with strong velocity folding (i.e. a low Nyquist velocity).
+#' @param x An object of class `pvol` or `scan`
+#' @param vp An object of class `pvol`, typically a vertical profile estimated for the input polar volume specified under `x` using `bioRad::calculate_vp()`.
+#' @param ... Currently not used.
+#' @param range_min,range_max Numeric. The minimum and maximum range to include, in m. Values are taken from the `vp`
+#' object if provided.
+#' @param alt_min,alt_max Numeric. The minimum and maximum altitude to include, in m. If only `alt_min` value is provided
```
adokter: the function errors with uninformative error message when selecting a layer outside the altitude range where profile data is available, e.g. alt_min=8000, alt_max=8500

R/vad.R
```diff
@@ -0,0 +1,264 @@
...
+#'
+#' Poor velocity fits in vertical profiles can arise in cases where the movement is not well described by a unidirectional
+#' velocity model, or for data with strong velocity folding (i.e. a low Nyquist velocity).
+#' @param x An object of class `pvol` or `scan`
+#' @param vp An object of class `pvol`, typically a vertical profile estimated for the input polar volume specified under `x` using `bioRad::calculate_vp()`.
+#' @param ... Currently not used.
+#' @param range_min,range_max Numeric. The minimum and maximum range to include, in m. Values are taken from the `vp`
+#' object if provided.
+#' @param alt_min,alt_max Numeric. The minimum and maximum altitude to include, in m. Separate panels will be plot for each
+#' altitude layer of the profile `vp`. If only `alt_min` is provided
```
adokter, line 15: when providing a single alt_min value, the plot doesn't indicate the altitude bin, as it does when you plot multiple altitude bins. better to be consistent and always show the altitude range as a label on the top.

adokter: Related: when I specify a narrow altitude range, like c(700,810), the panels are labeled as 600-800 and 800-1000, while it should be 700-800 and 800-810

bart1: > Related: when I specify a narrow altitude range, like c(700,810), the panels are labeled as 600-800 and 800-1000, while it should be 700-800 and 800-810

As I'm in a workshop I don't have much time this week. A quick remark here is that the labeling in this case would not reflect the altitude range of the vp sin that is plotted, as is currently the case. I guess either way is imperfect but up to now I have been thinking about the altitude bins as those from the vp and not the height selection of the pvol data

R/vad.R
bart1: made an issue: https://github.com/adokter/bioRad/issues/746

R/vad.R
bart1: Here the user can specify the reflectivity quantity to use for filtering. It is just that the default filtering works as described. If you use a specific quantity you can use an argument like ` range_gate_filter = TH>-10 & TH<20 & RHOHV<.95`. The reflectivity quantity is not used somewhere else in the function

R/vad.R
bart1: Have done this, called `velocity_quantity`, it seems an overkill to use tidy select here.

R/vad.R
bart1: I think you then you would still need several extra arguments like `reflectivity_quantity` and `rho_quantity`. Maybe a personal preference but I tend to want to keep the number of arguments limited. Using explicit thresholds would also reduce the option to use other filters (e.g. if you have a bird classifier or use differential reflectivity). I think for a function mostly useful for exploration this flexibility is valuable

R/vad.R
bart1: Because this calculates the angle above which larger then 5% underestimations of speed do occur (as in warning message). I will make a comment about it

R/vad.R
bart1: That is done when you use `cosine_correction="range_gates"`, then the points/range_gates are corrected. Here I adjust the sine curve. What could be improved here I now see is take the mean elevation angle per panel/height bin that we plot. Alternatively you would have to include a sine for each elevation angle that occurs in a height bin but that gets messy. For those small difference I would prefer one sine curve. Alternatively you use `cosine_correction="range_gates"` to avoid this averaging

R/vad.R
bart1: Not having looked into the vertical velocity estimates very much I do wonder if this would not add more noise then it resolves. I see the estimates are considered unreliable (`?summary.vp`) and sometimes are quite high (e.g. -15 m/s in `example_vp`). On the other hand even these high speeds will not change the sin very much for moderate elevation angles. I will look into this

R/vad.R
bart1: Good point, what to do when no range gate data gets selected. We could error however that can be annoying when using the function in a loop the alternative is to return an empty plot

R/vad.R
bart1: your probably right the current name is unfortunate. However I would prefer to also keep a version that does not do changes (for example as the corrections might be incorrect for nyquist folded data). So I need to think about the best way to do that with an `align_scans` argument.

R/vad.R
adokter: agree - throwing a warning and returning an empty plot sounds good

R/vad.R
adokter: Right, because the angles are so low high vertical velocities have very little effect on the fit. I just think it would be cleaner to show the actual fit instead of something else. Or have an option to remove vertical velocity in the plot

R/vad.R
adokter: Even if it's a little messy, I feel its more accurate than a mean, which we would have to explain carefully. You could try different colors for the points and sine curves of different sweeps, maybe that looks cleaner?

R/vad.R
adokter: I see your point, but it's very hard to understand for users what this does and how to modify it - and therefore prone to errors if people do try to change this. Could this be included as a customizable filter function to be included in the package, for which you can change the reflectivity and rho quantities and thresholds? There should be a way to simplify this
```
dplyr::if_all(utils::head(dplyr::matches(c("^DBZ$", "^DBZH$", "^DBZV$", "^TH$", "^TV$")), 1),
                         \(dbz) dbz_to_eta(dbz, !!x$attributes$how$wavelength) < 36000
                       ) &
                         dplyr::if_any(dplyr::matches("RHOHV"), \(rhohv) rhohv < .95)
```

PR #776, opened 2026-06-23
run devtools::document()

YohanSassi:


DESCRIPTION
```diff
@@ -75,3 +75,4 @@ VignetteBuilder: knitr
 Roxygen: list(markdown = TRUE)
 Config/testthat/edition: 3
 Config/roxygen2/version: 8.0.0
+RoxygenNote: 7.3.3
```
peterdesmet, line 78: I think Roxygen 8.0.0 is already used. @YohanSassi Can you try to update to that version?

peterdesmet (changes requested):

PR #785, opened 2026-07-29
add vertical Integration clean_mixture() quantities

adokter:
Adds vertically integrated quantities with function integrate_profile() whenever clean_mixture quantities are added. Requires application of `clean_mixture()` with option `keep_mixture=TRUE`


adokter:
fix needed: only use eta_mixture values in the height-integration for which the associated ground speeds are not NaN.

PR #789, opened 2026-08-11
Migrate ppi data to terra

iskandari:
- migrates `ppi$data` from `SpatialGridDataFrame` to `terra::SpatRaster` across PPI creation and downstream methods
- adds unit tests


bart1:
@adokter / @iskandari Quick idea would it be an option to use the terra object directly/using class inheritance. That would make all terra/spatial functions work directly on the results of a project_as_ppi function instead of always having to address a slot in the `ppi` object

